Issue #23, opened 2016-11-01
composer-install-elsewhere requires composer be installed globally


becw:
See https://github.com/palantirnet/the-build/blob/master/tasks/lib/artifacts.xml#L78

becw:
This still affects the release-2.0 branch.

becw:
Eep, there were some issues with this that are addressed in PR #115.

becw:
This work is included in the [2.0.0-alpha3 release](https://github.com/palantirnet/the-build/releases/tag/2.0.0-alpha3).

PR #34, opened 2016-11-21
Fix dependency on global composer

becw:
I don't think this is the right way to solve this. Instead, we should use the [ComposerTask](https://www.phing.info/docs/master/hlhtml/index.html#ComposerTask), which allows you to set the path to composer if necessary (we could set it up to work like the DrushTask, which allows you to set the `drush.bin` property).

See issue #23.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,11 +9,12 @@
         }
     ],
     "require": {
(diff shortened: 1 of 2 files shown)


becw:
Fixed in release-2.0.

PR #47, opened 2017-01-12
Add memcache config for Drupal

becw:
The problem here is that you can only enable the memcache cache backend after Drupal is installed... and when the memcache module is present and enabled. So these settings.php changes will cause problems on Acquia if memcache isn't present, and even locally you have to manually edit the `settings.php` file after site installation to use memcache, which is not going to be a consistent way of enabling that config.


becw:
This should be updated to use Acquia's recommended default file instead, which is linked here: https://docs.acquia.com/acquia-cloud/performance/memcached/intro/

becw:
I'm going to close this in favor of #141, since this PR is now very out of date and isn't the approach we want to use anyway.

PR #61, opened 2017-02-09
Fail if deleting files fails.

becw:
To test:
- maybe you can checkout this branch in your vendor/palantirnet/the-build
- and then run your build again and see if it still fails this way?

### DIFF
--- a/tasks/lib/artifacts.xml
+++ b/tasks/lib/artifacts.xml
@@ -99,7 +99,7 @@
         <!--
             then we composer install in the build dir...
             -->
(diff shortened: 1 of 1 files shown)


zendoodles:
Agree with @liberatr. This did not solve the problem for me.

Also, @becw it might also be good to know why these file permissions are broken in the first place. Is that something I've done wrong somewhere I could stop doing?

becw:
I'm not going to pursue this fix on the 1.0 branch.

PR #64, opened 2017-02-14
Composer remove dev dependencies

becw:
followup to "fail if deleting files fails" (#61).


becw:
I'm not going to pursue this change on the 1.0 branch.

PR #70, opened 2017-03-15
Multi site config, build and install

zendoodles:
## Description:

Adds multisite configuration, build and install tools to the-build:

Still needs some work for deploy artifacts and drush.

The drushrc.php method assumes a single URI, so it is impossible to install multiple sites with the-build because no matter how well you've configured, it always uses the default uri. Need to decide how to proceed there.

Deployment makes some similar assumptions. I *think* I've resolved the bulk of it, but it still needs a bit more work and a lot of poking to see what breaks.

### `vendor/bin/phing configure` 
Behaves as before. It creates .properties files which will be used by the build script below. Configuration for environment is separated from that of the site itself. This allows us to use any number of combinations of environments and sites without files for every permutation of the two. So for example, on most projects we would have:
* `build.vagrant.env.properties` for environment specific settings like the list of sites to build for this environment (build.sites), artifact mode (build.artifact_mode), and the drupal root (build.drupal.root). In cases where there will not be a multisite environment, we still use this configuration, but there is only one site, "default" in the build.sites list.
* `build.default.site.properties` for the settings should be used in sites/default/settings.php.

### `vendor/bin/phing drupal-build`
Builds all sites configured for this environment by looping thru build.sites and calling prepare-site on each one. Assuming configuration is built with this version, this is safe to use as before even if there is only one site. One is a subset of many.

### `vendor/bin/phing drupal-install-site`
Installs an individual site for a given environment. You can pass the name of the site to build like `vendor/bin/phing drupal-install-site -Ddrupal-site=default` otherwise it will ask which site to install.

### `vendor/bin/phing drupal-install`
Locally installs all the Drupals configured for the chosen environment by looping thru build.sites and calling `drupal-install-site` on each one. This is also safe to use as before even when there is only one site configured. One is a subset of many.

## To Test:
- [ ] Check out [the companion PR for the AMA d8 environment](https://github.com/AmericanMedicalAssociation/ama-d8/pull/4) and bring up the VM.
- [ ] Run `vendor/bin/phing build install` and validate two sites are installed.
- [ ] Run the behat tests and note they are being run against both environments.
- [ ] Configure a new project using the drupal-skeleton and this branch. Use the `vendor/bin/phing configure` command and validate it writes the configuration files as expected based on the instructions above.

## Automated Tests:

* https://github.com/AmericanMedicalAssociation/ama-d8/blob/EWL-3131_Multisite/features/test.ama-d8.local/installation.feature
* https://github.com/AmericanMedicalAssociation/ama-d8/blob/EWL-3131_Multisite/features/www.ama-d8.local/installation.feature
---

[Pull Request Guidelines](https://github.com/palantirnet/development_documentation/blob/master/guidelines/pr_review.md)

### DIFF
--- a/tasks/acquia.xml
+++ b/tasks/acquia.xml
@@ -88,7 +88,6 @@
             <property name="install_dir" value="${repo.dir}" />
             <property name="build.dir" value="${build.dir}" />
             <property name="drupal.root" value="${drupal.root}" />
-            <property name="drupal.sites_subdir" value="${drupal.sites_subdir}" />
         </phing>
     </target>
 
--- a/tasks/boilerplate.xml
+++ b/tasks/boilerplate.xml
@@ -12,6 +12,7 @@
       Include this file in your build.xml with:
         <import file="vendor/palantirnet/the-build/tasks/boilerplate.xml" />
       -->
+    <property name="build.sites" value="default" />
 
     <taskdef name="includeresource" classname="TheBuild\IncludeResourceTask" />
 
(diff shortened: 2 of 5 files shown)


becw:
Fixed in release-2.0.

PR #76, opened 2017-05-26
Download backups from Acquia Cloud

becw:
```xml
<DownloadLatestAcquiaBackup dest="artifacts/databases" realm="prod" site="massdocs" env="prod" conf="artifacts/acquiacloud/.acquia/cloudapi.conf" resultProperty="db.load.file" />
```

Wait a second... there are already Acquia db download tasks.


becw:
Fixed in release-2.0.

PR #82, opened 2017-08-03
Require forked version of phing-drush-task

froboy:
Drupal 8.4 requires Symphony 3 which isn't currently supported in Drush 8. See https://github.com/drush-ops/drush/pull/2724 for more. phing-drush-task is locked to drush < 8.1, so I forked it and changed composer.json, but I ran into some composer issues getting the forked library to load. Instead of trying to debug PSR-0 I just rewrite the handful of phing drush commands in the build with straight exec commands which run in the directory set by the build variables.

This also removes `php-todo-finder` https://github.com/nilportugues/php-todo-finder/blob/master/composer.json as it's requiring old Symfony libraries and and doesn't seem to be used.

@becw Drupal 8.4 is not installable with our build tools until this or a similar fix is accepted.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,22 @@
     "bin": [
         "bin/the-build-installer"
     ],
+    "repositories": [
+        {
+          "type": "package",
+          "package": {
+              "name": "continuousphp/phing-drush-task",
+              "version": "dev-update-drush",
+              "source": {
+                  "url": "https://github.com/froboy/phing-drush-task.git",
+                  "type": "git",
+                  "reference": "update-drush"
+              }
+          }
+        }
+    ],
     "require": {
+        "continuousphp/phing-drush-task": "dev-update-drush as dev-master",
         "drupal/coder": "*",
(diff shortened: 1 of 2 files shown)


becw:
An alternative to this approach is to alias the version of Drush required by `palantirnet/drupal-skeleton`. Since the dependency issue is ultimately a conflict between the dependencies of that repository, I think that approach introduces fewer moving parts/things to update later. Here's how that change looks: https://github.com/palantirnet/drupal-skeleton/pull/70

becw:
I've created a new branch based on this work - see PR #85. I found that:

* Command syntax has changed for some commands in Drush 9
* When removing the `nilportugues/php_todo` library, I needed to remove the associated functionality
* Updating the `continuousphp/phing-drush-task` library for Drush 9 was pretty straightforward

becw:
Also, it turns out this update breaks backward compatibility!

becw:
The remote branch has been merged, so this should no longer be an issue.

I'm going to leave the branch in case anyone was using it.

PR #86, opened 2017-10-30
Empty contents of files and private files directories, rather than de…

becw:
…leting, to allow using this task on read-only environments (e.g. platform.sh).


becw:
This turned out to be kind of a pain in certain circumstances where we wouldn't want to accidentally delete files.

PR #91, opened 2018-07-13
Artifact rearchitect

becw:
Rewrite the artifact build process so that:

* there's a generic implementation
* it doesn't have as many hidden properties
* it can build to multiple branches

This is totally in progress and will break any existing artifact build processes.

This is intended to be more reusable and understandable than the previous version. That doesn't mean that it's simpler -- but:

* Code flow should be easier to follow
* Output should be more helpful
* Functionality should be more reusable with fewer customizations per project

### Remaining tasks

- [ ] Update path from previous acquia artifact builds
- [x] update properties documentation
- [x] set default artifact path in the code
- [x] update properties documentation in the code


tasks/artifact.xml
```diff
@@ -0,0 +1,318 @@
...
+             to avoid branch conflicts. -->
+        <property name="artifact.git.temporary_branch" value="artifact-${artifact.git.commit}" />
+        <!-- Prefix the repository tag so that we're not using the exact same tag on the
+             artifact and on the repository, to avoid confusion, especially when the
+             artifact is built on a branch of the development repository. -->
+        <property name="artifact.git.artifact_tag" value="${artifact.prefix}${artifact.git.tag}" />
+        <!-- If the remote branch isn't configured, use a remote branch based on the name
+             of the current branch. -->
+        <if>
+            <not><equals arg1="${artifact.git.remote_branch}" arg2="" /></not>
```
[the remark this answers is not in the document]
becw: Oh dang, nice logic catch -- the intent was to set the `artifact.git.remote_branch` if it was empty, so that the artifact could automatically build to a remote branch that corresponds with the local branch (e.g. `artifact-PNET-123` or `artifact-fix-sidebar`).

... But as you point out, it will not do that.

tasks/artifact.xml
```diff
@@ -0,0 +1,314 @@
...
+            <fileset dir="${artifact.directory}" defaultexcludes="false" excludes=".gitignore,.git/,.git/**" includes="**/.gitignore,**/.git,**/.git/**" />
+        </delete>
+
+        <echo>Running 'phing build' in the artifact...</echo>
+        <!--
+            - Override the Drupal root so that it is within the artifact directory.
+            - Override the "artifact mode" so that files managed with the
+              <includeresource /> task are copied into the artifact instead of symlinked.
+            -->
+        <phing target="build" phingfile="${phing.file}" inheritAll="false" dir=".">
```
byrond: Would `inheritAll` keep `build.env` from being passed here?  When I tested this, the artifact was built with a `settings.php` for The Vagrant instead of Acquia's.

byrond: Never mind.  I didn't have an `build.acquia.properties` that specified the Acquia settings file.

tasks/artifact.xml
```diff
@@ -0,0 +1,314 @@
...
+  Copyright 2018 Palantir.net, Inc.
+  -->
+
+<project name="artifact" default="artifact-main">
+
+
+    <!--
+      Required configuration for building artifacts:
+        artifact.directory
+        artifact.git.remote
```
byrond: Should this have a description?  It wasn't initially clear to me that it was the full URL to the Git repository on the host.

byrond:
This worked great, @becw! I tested on an Acquia Cloud Free site with a fresh project that used this branch of The Build. I tried most of the options available to `phing artifact`, and they worked perfectly.

becw:
I'm going to close this PR, and merge this work into the `release-2.0` branch, which already contains most of it. Since this removes some of the other artifact stuff, this as a backwards compatibility breaker. I think we're ready to move forward based on this PR convo, though!

PR #93, opened 2018-08-23
Add styleguide build command.

becw:
Completes [ZRAD-119](https://palantir.atlassian.net/browse/ZRAD-119) - Add styleguide target to palantirnet/the-build.

### To test:

1. Require this branch of the-build in a project: `composer require --dev palantirnet/the-build:dev-styleguide-build-commands`
2. Reinstall the-build: `vendor/bin/the-build-installer`
3. When prompted, make sure that the styleguide directory and styleguide command are correct for your styleguide (if you miss the prompts, just re-run the installer)
4. Run the build command: `phing build` -- this should install the styleguide dependencies and run the styleguide command that you configured
5. Run the styleguide command directly: `phing styleguide` -- since your dependencies are already installed, this should just rerun the styleguide command

### Open questions:

* Will this cover most of our styleguide use cases? Folks working on the styleguide will continue to use the yarn/npm gulp-based tooling for development. The intent of this command is to generate the assets required for an artifact build.
* This only supports composer and yarn right now -- and will only run install commands if those lockfiles are present. Do we need to also support npm?
  * Are we back to using npm only in some styleguides?
* Where should we document that this can be configured on existing projects by:
  1. Adding the new `<import>` line for the styleguide targets to the project's `build.xml`
  2. Adding the phing call to `styleguide` to the existing `build` target, if desired
  3. Configuring the styleguide variables with `phing styleguide-configure -Dbuild.env=default`

### DIFF
--- a/build.dist.xml
+++ b/build.dist.xml
@@ -14,8 +14,11 @@
 
     <!-- Default target: build -->
     <target name="build" description="Build the application.">
-        <phingcall target="drupal-build" />
         <phingcall target="styleguide" />
+        <phingcall target="drupal-build" />
+        <!-- Include styleguide resources in the theme. This approach will symlink
+             resources in development environments, and copy them for artifact builds. -->
+        <!-- <includeresource mode="${build.artifact_mode}" source="styleguide/source/assets/css" dest="${drupal.root}/themes/custom/example_theme/css" /> -->
     </target>
 
 
--- a/build.dist.xml
+++ b/build.dist.xml
@@ -9,6 +9,7 @@
     <!-- Make these commands available by default. -->
     <import file="vendor/palantirnet/the-build/tasks/drupal.xml" />
     <import file="vendor/palantirnet/the-build/tasks/acquia.xml" />
+    <import file="vendor/palantirnet/the-build/tasks/styleguide.xml" />
 
 
     <!-- Default target: build -->
--- a/build.dist.xml
+++ b/build.dist.xml
@@ -15,6 +15,7 @@
     <!-- Default target: build -->
     <target name="build" description="Build the application.">
         <phingcall target="drupal-build" />
+        <phingcall target="styleguide" />
     </target>
 
 
--- a/tasks/styleguide.xml
+++ b/tasks/styleguide.xml
@@ -85,18 +85,6 @@
                 <exec dir="${styleguide.root.resolved}" command="yarn install --non-interactive --no-progress --prefer-offline" passthru="true" checkreturn="true" />
             </then>
         </if>
-
-        <!-- Run npm install in the styleguide, if there's no yarn.lock file. -->
-        <!-- @todo is there a decent way to check whether npm install has run? -->
-        <if>
-            <and>
-                <available file="${styleguide.root.resolved}/package.json" type="file" />
-                <not><available file="${styleguide.root.resolved}/yarn.lock" type="file" /></not>
-            </and>
-            <then>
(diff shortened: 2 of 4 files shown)


build.dist.xml
```diff
@@ -9,11 +9,13 @@
     <!-- Make these commands available by default. -->
     <import file="vendor/palantirnet/the-build/tasks/drupal.xml" />
     <import file="vendor/palantirnet/the-build/tasks/acquia.xml" />
+    <import file="vendor/palantirnet/the-build/tasks/styleguide.xml" />
 
 
     <!-- Default target: build -->
     <target name="build" description="Build the application.">
         <phingcall target="drupal-build" />
+        <phingcall target="styleguide" />
```
byrond: I'm testing this on a project that uses the `copy` instead of `symlink` artifact mode.  Does it make sense to move the `<phingcall>` above the build, so the styleguide assets are updated before being copied into Drupal's theme?

byrond:
Running this by default in the VM will cause issues for devs using Yarn on their host machine.  When there's a mismatch, it results in the following type of error:
```
Node Sass could not find a binding for your current environment: OS X 64-bit with Node.js 6.x

Found bindings for the following environments:
  - Linux 64-bit with Node.js 6.x
```
Would it be best to limit this to `build.circle.properties` by default?

byrond:
For documentation, it might be a good idea to add a section to the project readme, with maybe a link to a new document in `docs/` with details on updating existing projects to use the new styleguide commands.  Also, it would probably be good to add the new properties to `docs/properties.md`.

build.dist.xml
russom-woldezghi: @byrond Yeah that makes sense to have a copy/updated before being copied over into Drupal's theme. Maybe with an output of what is copied over to help users debug - if there is a need to troubleshoot/debug changes.

russom-woldezghi:
> For documentation, it might be a good idea to add a section to the project readme, with maybe a link to a new document in `docs/` with details on updating existing projects to use the new styleguide commands. Also, it would probably be good to add the new properties to `docs/properties.md`.

We could also point to have users run: `vendor/bin/phing configure -Dbuild.env=your_environment` to get all the possible options?

build.dist.xml
becw: Thanks! I've made this change, and I've added an example of how to include theme assets using tools provided by the-build.

tasks/styleguide.xml
```diff
@@ -0,0 +1,91 @@
...
+
+            <property name="update" value="styleguide.root,styleguide.command" />
+            <property name="dump" value="" />
+        </phing>
+    </target>
+
+
+    <!-- Target: styleguide -->
+    <target name="styleguide" description="Install and build the style guide.">
+
```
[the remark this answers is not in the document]
becw: That's a good point -- especially re: the nodejs installs that vary by platform.

The thing is, where the styleguide is built tends to vary between development roles -- folks doing design and front end development will build it locally, but folks doing engineering, devops, or support will build the styleguide within the VM.

So I'm not sure where to go with this...

byrond:
Requires https://github.com/palantirnet/butler/pull/101 and https://github.com/palantirnet/the-vagrant/pull/56 for new projects.

becw:
Since we're not using Butler for new projects, I'm not going to worry about https://github.com/palantirnet/butler/pull/101 :)

becw:
... and, I'm going to close this in favor of #117, which I just merged! Thank you, @byrond and @russom-woldezghi, for uncovering and pursuing these issues! Your work in this PR made it super easy to update this work for the release-2.0 branch.

PR #94, opened 2018-08-23
Remove the 'drupal.modules_enable' property/behavior

becw:
Remove the 'drupal.modules_enable' property -- sites should use config_split instead.


becw:
Oh, that's a good idea.

becw:
@byrond I've added deprecation messages -- is that text helpful, or is there some other info I should add?

PR #95, opened 2018-08-24
Allow specifying a node version on CircleCI.

becw:
Open questions:

* Should we cache the `~/.nvm` directory?
* Should this be uncommented by default?


becw:
Ooh, interesting... I've made the changes you suggested.

becw:
I tested these changes over on the UCL repo and it works exactly as you describe :)

PR #96, opened 2018-09-14
Expose code review commands in the template build.xml

becw:
This change removes the individual phing XML files for each code review task, and moves the reviews to be explicitly invoked in the template `build.xml` file.

Removing the buried code should make it easier for folks to tell what code reviews are running on their project, and to customize them instead of just disabling them when they're not working.

This change is against the `release-2.0` branch, and I have no plans to backport it.


tasks/review.xml
```diff
@@ -0,0 +1,56 @@
...
+
+<project name="review" default="review-properties">
+
+
+    <!--
+      PHP Mess Detector configuration.
+      These values are used in the build.dist.xml template:
+        phpmd ${phpmd.directories} ${phpmd.format} ${phpmd.rulesets} -.-suffixes=${phpmd.suffixes}
+      -->
+    <property name="phpmd.directories" value="${drupal.root}/modules/custom/,${drupal.root}/themes/custom/" />
```
byrond: Should `profiles/custom` be included by default, since we are linting them by default?

tasks/review.xml
```diff
@@ -0,0 +1,56 @@
...
+    <property name="phpmd.format" value="text" />
+    <property name="phpmd.rulesets" value="${build.thebuild.dir}/conf/phpmd.xml" />
+    <property name="phpmd.suffixes" value="php,inc,module,install,profile" />
```
byrond: Does this also need to check `.test` and `.theme` files by default?

tasks/review.xml
```diff
@@ -0,0 +1,56 @@
...
+
+
+    <!--
+      PHP Code Sniffer configuration.
+      These values are used in the build.dist.xml template:
+        phpcs -.-standard=${phpcs.standard} -.-ignore=${phpcs.ignore} ${phpcs.directories}
+      -->
+    <property name="phpcs.standard" value="${build.dir}/vendor/drupal/coder/coder_sniffer/Drupal/ruleset.xml" />
+    <property name="phpcs.directories" value="${drupal.root}/modules/custom ${drupal.root}/themes/custom" />
```
byrond: Should profiles/custom be included by default, since we are linting them by default?

byrond:
One more thing.  I assume the 2.0 branch won't support Drupal 7, and that's the reason for removing the D7 file patterns from the reviews.

becw:
Right on, thank you @byrond!

Yes to all of the places where you pointed out that the files under review are inconsistent -- I'll fix those up.

And, yes -- I think the 2.0 branch should drop support for Drupal 7, especially since it might make sense to update to Drush 9 soon, which doesn't support d7 or early versions of d8.

PR #98, opened 2018-09-25
Consolidate defaults

becw:
* Consolidates default values into `defaults.properties.yml`
* Reorganizes documentation to reflect consolidated property defaults
* Moves `boilerplate.xml` --> `the-build.xml` to more accurately reflect the role that file plays
* Removes unused targets and files


defaults.properties.yml
```diff
@@ -1,30 +1,192 @@
...
+    # This property is commented out, because the value should only be set in per-project
+    # properties files.
+    #remote:
+
+    # Name of the remote branch to push the artifact to.
+    # The remote_branch property is generally set dynamically, to 'artifact.prefix' plus
+    # the current branch name. However, it may be overridden either in your project
+    # properties or in a run-time flag (-Dartifact.git.remote_branch=foo) in order to
+    # build to a static destination branch, or a branch name that matches other
+    # constraints (like Pantheon's 11 character limit for multidev environment branch
```
byrond: Great example!

tasks/the-build.xml
```diff
@@ -0,0 +1,77 @@
...
+
+
+    <!-- Alias the build directory to the Phing start directory. -->
+    <property name="build.dir" value="${application.startdir}" />
+
+    <!-- Use the project directory name as the project name -->
+    <basename property="projectname" file="${build.dir}" suffix="local" />
+
+    <!-- Provide the path to the-build's code for loading templates and defaults -->
+    <!-- Phing maps the project name "the-build" to "the_build". Should we use the_build instead to avoid this confusion? -->
```
byrond: Sticking with `the-build` seems good to me.  It's pretty common for the dash/underscore mapping to happen, and sticking with the dash seems like it may be better for URLs and such.

tasks/the-build.xml
```diff
@@ -0,0 +1,77 @@
...
+            </if>
+
+            <!-- Load the project's defaults. This may be loaded twice, but that
+                 should not be a problem. -->
+            <property file="${build.dir}/.the-build/build.default.properties.yml" />
+
+            <!-- Load the-build's global defaults. -->
+            <property file="${build.thebuild.dir}/defaults.properties.yml" />
+
+            <!-- Prevent the-build from reloading properties. -->
```
byrond: Interesting.  Does this get loaded/run more than once?

defaults.properties.yml
becw: This was totally from my work with Avi on NTC -- we were so confused until we figured out Pantheon's branch naming restrictions :(

defaults.properties.yml
```diff
@@ -1,30 +1,192 @@
-build:
-  artifact_mode: symlink
```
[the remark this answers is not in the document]
becw: I think that the only place that it really makes sense to copy instead of symlink is when building the artifact, which is done inline -- so I thought that relying on the class behavior/defaults makes sense instead of treating this as a configuration property.

tasks/the-build.xml
becw: IIRC, in some cases we've run targets using [PhingTask](https://www.phing.info/phing/guide/en/output/hlhtml/#PhingTask), which loads up the phing project and runs a target (vs. [PhingCallTask](https://www.phing.info/phing/guide/en/output/hlhtml/#PhingCallTask), which runs a target in the current project). I think we've used this approach in past artifact builds (for building but overriding particular properties, like `drupal.root`) and for multisite installation (overriding `drupal.uri` and `drupal.database.database`).

This approach is still used in [install.xml](https://github.com/palantirnet/the-build/blob/21fd7c5625ffaaaf9d4c8ba0cb6b60eddc085953/tasks/install.xml#L191-L192) and [artifact.xml](https://github.com/palantirnet/the-build/blob/21fd7c5625ffaaaf9d4c8ba0cb6b60eddc085953/tasks/artifact.xml#L262-L265), but with the `inheritAll="false"` parameter, so I'm not sure whether this check is relevant... but I'm reluctant to remove it.

tasks/lib/repositories.xml
```diff
@@ -1,253 +0,0 @@
...
-                            </catch>
-                        </trycatch>
-                    </then>
-                </if>
-            </else>
-        </if>
-    </target>
-
-
-    <target name="commit">
```
becw: Oops! Looks like this target is still used by [artifact.xml](https://github.com/palantirnet/the-build/blob/21fd7c5625ffaaaf9d4c8ba0cb6b60eddc085953/tasks/artifact.xml#L269-L276). I better fix that!

becw:
I made a couple of fixes related to the artifact -- I was testing the removed `repositories.xml` file, and there were some issues with the artifact build.

tasks/the-build.xml
byrond: Good point.  Definitely safe to keep it in. 👍

defaults.properties.yml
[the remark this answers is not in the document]
byrond: Sounds good to me.  That helps to simplify things a bit.

defaults.properties.yml
byrond: 👍

PR #99, opened 2018-09-27
Flatten the Drupal build tasks

becw:
* Moves the `drupal-build` and `drupal-install` targets into the template `build.xml` where developers can see and modify them
* Moves the db loading target into `drupal.xml`, and makes it available in the template `build.xml` file
* Makes the Acquia targets available in the template `build.xml` file by default
* Updates properties used in the `drupal.xml` file so that they are all namespaced with `drupal.`

This PR follows #98.


defaults/templates/the-build/build.xml
```diff
@@ -1,25 +1,75 @@
...
     <target name="build" description="Build the application.">
-        <phingcall target="drupal-build" />
+        <!-- Create the Drupal custom code directories. -->
+        <foreach list="${drupal.create_dirs}" param="dir" target="mkdir" />
+
+        <!-- Set permissions on the sites subdirectory. -->
+        <mkdir dir="${drupal.root}/sites/${drupal.sites_subdir}" mode="750" />
+
+        <!-- Create the public files directory and make it world writable. -->
+        <mkdir dir="${drupal.root}/${drupal.settings.file_public_path}" mode="666" />
```
byrond: Drupal was throwing permission errors when trying to create directories here without mode `777`.  Will the `.htaccess` files that get created her and in the private files directory keep scripts from being executed?

tasks/drupal.xml
```diff
@@ -21,142 +21,68 @@
...
+    <target name="drupal-validate-clean-config" description="Validate clean drupal config directory.">
+      <!-- Assembling this property at run time. -->
+      <property name="drupal.validate_clean_config.dir" value="${build.dir}/${drupal.root}/${drupal.config_sync_directory}" />
 
-    <!-- Target: create-placeholder -->
-    <target name="create-placeholder">
-        <fail unless="placeholder_for" message="The 'placeholder_for' property is required." />
-        <resolvepath propertyName="placeholder_for.resolved" file="${placeholder_for}" dir="${build.dir}" />
+        <!-- Use git to list local (not yet committed) changes to the config directory. -->
+        <exec command="git status --porcelain ${drupal.validate_clean_config.dir}" outputProperty="modified_files" />
```
byrond: I encountered a strange situation where, when the `config` directory doesn't exist, the build fails here.  However, using `bypass` doesn't display the `modified_files`, implying that it is indeed empty.  So, I would expect that if it fails here, it should print the message "Continuing install with these local changes..." when using `bypass`.

I only ran into this because I didn't install Drupal as part of `the-build-installer`.  I was following along with the Drupal Skeleton readme, which instructs running the installer outside the VM before `vagrant up`.  Do you think we should change that when this is merged, or is there a way we can make `phing install` do the initial install if it has not been already?

tasks/acquia.xml
```diff
@@ -19,25 +19,33 @@
     <fail unless="build.dir" />
     <fail unless="build.env" />
 
-    <fail unless="acquia.accountname" />
```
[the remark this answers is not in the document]
becw: Because I've added the `acquia.xml` tasks to the `build.xml` template -- so including these lines at the top level of this file would mean that the `acquia.accountname` property would have to be configured in order to run any phing task.

Users would have to know to go and manually set this property in their config -- and not everybody will use these tasks anyway.

So, moving these checks back inside the specific targets made sense because this way, they only warn folks who run those targets without all of the required properties configured.

tasks/drupal.xml
becw: Oh, that's an interesting idea -- to offer to run the first-install target if Drupal hasn't been installed yet. I've been wondering how to handle this, since I don't necessarily expect folks to install the-build from within a Vagrant environment...

becw: Ok, I've tried to address this issue! I've made it so that, if there's no config to install from, the-build offers to install a fresh Drupal when you run `phing install`. This didn't involve changing anything else about the `drupal-validate-clean-config` target, but let me know if that part still gets in your way.

becw:
@byrond I rebased this on top of the rest of the change from #98, sorry if that trips you up! Otherwise, I've addressed the issues you ran into -- I think. Can you please test again?

tasks/acquia.xml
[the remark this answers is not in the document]
byrond: Oooooh.  Good point!

tasks/install.xml
```diff
@@ -180,20 +181,31 @@
...
-                <input propertyName="install_now" validArgs="y,n" message="Install Drupal now" promptChar="?" />
+        <!-- Run the first build. -->
+        <phing dir="${application.startdir}" target="build" inheritAll="false" />
+
+        <!-- If Drupal has a database connection available, offer to install Drupal. -->
+        <trycatch>
+            <try>
+                <!-- The 'haltonfailure' attribute seems to have the opposite behavior
+                     from what is described in the Phing documentation. -->
+                <phing dir="${application.startdir}" target="drupal-has-database-connection" inheritAll="false" haltonfailure="true" />
```
byrond: When running `the-build-installer` outside the VM, it now fails with "Drupal database connection not available" because of the try/catch inside the `drupal-has-database-connection` task.

defaults/templates/the-build/build.xml
```diff
@@ -1,25 +1,78 @@
...
+
+        <!-- Create the public files directory and make it world writable. -->
+        <mkdir dir="${drupal.root}/${drupal.settings.file_public_path}" />
+        <chmod file="${drupal.root}/${drupal.settings.file_public_path}" mode="744" />
+
+        <!-- Create the private files directory and make it private. -->
+        <mkdir dir="${drupal.root}/${drupal.settings.file_private_path}" />
+        <chmod file="${drupal.root}/${drupal.settings.file_private_path}" mode="640" />
+
+        <!-- Copy the settings.build.php and services.build.php templates into place. -->
```
byrond: I'm not sure if it's happening in this section, but I was testing an artifact build with Pantheon and noticed `settings.build.php` is getting added, which breaks the site on Pantheon.  We're calling this from `artifact-build` with `inheritAll="false"` which makes me think maybe `build.drupal.settings` et. al. are getting reset to defaults.  Do we need to pass them in from `artifact-build`?

byrond: Along those lines, it might be good to add a `pantheon.yml` that sets `web_docroot=true`.  I'm not sure where it's best to add, though.  Maybe in the installer if Pantheon is selected as the hosting platform?

byrond:
@becw I opened https://github.com/palantirnet/drupal-skeleton/pull/77 against the skeleton to test on CircleCI and update the config for these changes.

defaults/templates/the-build/build.xml
becw: Ah, interesting. I would love to add Pantheon and Platform config but I was holding off because it would be easy to rabbit hole -- but that's exactly where I would add it, in the installer.

I do think the first bit (about settings.build.php being added to a pantheon artifact) needs to be fixed before this can be merged, so I'll check that out.

tasks/install.xml
becw: Oops, it shouldn't fail, it should just not do that part of the install. Let me check that out.

defaults/templates/the-build/build.xml
becw: Can you give me more details on this issue? What I find is that when I run `phing build -Dbuild.env=pantheon`, the-build does create a `settings.build.php` file, but it's a limited version of the settings, customized for Pantheon:

```php
<?php

/**
 * @file
 * Drupal settings file template for use on Pantheon environments.
 */

$config_directories = [];
$config_directories[CONFIG_SYNC_DIRECTORY] = '../config/sites/default';
```

Is it this file that causes issues? What issues does it cause?

tasks/install.xml
becw: I've added a message/output to clarify what's happening here -- it's just skipping the Drupal install, but the rest of the-build is already installed.

defaults/templates/the-build/build.xml
becw: Ohhh, wait, in the artifact process. Hang on...

becw: ... and I went ahead and added Pantheon and Platform.sh hosting config files.

becw:
@byrond I've done another round of revisions based on the issues you ran into. Can you please test again?

byrond:
@becw should `settings.acquia.php` and `settings.pantheon.php` be in this list?
https://github.com/palantirnet/the-build/blob/1d873373e82109786df82c8f0f422e244d9364b9/defaults/templates/project.gitignore#L10-L13

Issue #20, opened 2016-08-15
Better Twig debugging?


becw:
Implement the full setup from https://www.drupal.org/node/1903374 on development, and make sure that it doesn't carry through to prod?

froboy:
Long time, no see, friendo. I implemented this on AMA at https://github.com/AmericanMedicalAssociation/ama-d8/pull/44/commits/d4c4dc98624c72f5dc6953160569a91242920c8e

For those who can't see that repo, here's the diff:

```
From d4c4dc98624c72f5dc6953160569a91242920c8e Mon Sep 17 00:00:00 2001
From: Avi Schwab <schwab@palantir.net>
Date: Thu, 31 Aug 2017 17:27:10 -0500
Subject: [PATCH] Make sure twig debugging is really really on. as per
 https://www.drupal.org/node/2598914

---
 provisioning/templates/services.j2.yml | 11 ++++++++---
 provisioning/templates/settings.j2.php |  5 +++++
 2 files changed, 13 insertions(+), 3 deletions(-)

diff --git a/provisioning/templates/services.j2.yml b/provisioning/templates/services.j2.yml
index 55ebddb..4179b5a 100644
--- a/provisioning/templates/services.j2.yml
+++ b/provisioning/templates/services.j2.yml
@@ -6,9 +6,9 @@ parameters:
     gc_maxlifetime: 200000
     cookie_lifetime: 2000000
   twig.config:
-    debug: {{ item.twig_debug }}
-    auto_reload: null
-    cache: true
+    debug: {{ item.twig_debug|lower }}
+    auto_reload: {{ 'true' if item.twig_debug else 'null' }}
+    cache: {{ (not item.twig_debug)|lower }}
   renderer.config:
     required_cache_contexts: ['languages:language_interface', 'theme', 'user.permissions']
     auto_placeholder_conditions:
@@ -34,3 +34,8 @@ parameters:
     - sftp
     - webcal
     - rtsp
+{% if item.twig_debug %}
+services:
+  cache.backend.null:
+    class: Drupal\Core\Cache\NullBackendFactory
+{% endif %}
diff --git a/provisioning/templates/settings.j2.php b/provisioning/templates/settings.j2.php
index e2a5547..7a5b448 100644
--- a/provisioning/templates/settings.j2.php
+++ b/provisioning/templates/settings.j2.php
@@ -2,6 +2,11 @@
 
 $settings['cache']['default'] = 'cache.backend.database';
 
[...]

PR #107, opened 2018-10-17
Disable render cache for development

becw:
Addresses #20.

This file (`drupal.settings.build.php`) is only used on local and CI environments. Acquia, Pantheon, and Platform.sh environments use an entirely separate settings file template.


becw:
Awesome, thanks for checking this out! I followed the docs :)

PR #110, opened 2018-10-18
Allow configuring what command to use for running composer in the artifact

### LINKED ISSUE #23: composer-install-elsewhere requires composer be installed globally

becw:
Resolves #23 -- though it requires that you add a bit of configuration if you don't have the `composer` or `composer.phar` commands available globally.

### DIFF
--- a/defaults.properties.yml
+++ b/defaults.properties.yml
@@ -236,6 +236,7 @@ phpmd:
   # File extensions to review.
   suffixes: php,inc,module,theme,profile,install,test
 
+
 # Configuration for running behat tests.
 #
 # Putting these flags in configuration allows you to vary the behat configuration per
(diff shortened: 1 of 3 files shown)


defaults.properties.yml
```diff
@@ -248,3 +249,11 @@ phpmd:
...
 behat:
   args: "--suite=default --strict"
+
+
+# Configuration for composer.
+#
+# Use this configuration if you're running on an environment where composer is not available on the command line as
+# `composer` or `composer.phar`.
+# composer:
+#   composer: ${build.thebuild.dir}/vendor/bin/composer.phar
```
byrond: Would this require us to have `"composer/composer": "^1.7"` (or something similar) as a dependency for the-build (like https://github.com/palantirnet/the-build/pull/34/files)?  Also, I did some local testing of this method by moving `composer` out of my path in the VM, and I had to use `vendor/bin/composer` as the path in this config.

becw: I tweaked this description.

becw: Basically, you may or may not want to add composer to your project... if you're checking dependencies in to your repo, then sure, but otherwise you may just have it in a non-standard location.

byrond: That's helpful!  Thanks for clearing that up.

PR #113, opened 2018-11-08
Add a "code-fix" target to automatically fix code violations

patrickfweston:
The `code-review` task runs the PHP Code Sniffer which logs coding standard violations. It is possible to automatically fix these errors, but doing so requires running a long command. This PR adds a target called `code-review` to fix violations that PHPCBF can address automatically.

Resources:
- https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically#using-the-php-code-beautifier-and-fixer


becw:
😍

Issue #109, opened 2018-10-18
Allow keeping the result of an artifact build without cleaning it up


becw:
Currently, the end of the artifact target prompts the user asking to push the artifact. If you say no, it will automatically clean up all changes to the artifact repo.

You can go in and look at the changes while the prompt sits there, if you like -- but that's not intuitive.

Instead, it should allow you to push, not push but keep the artifact, or discard the changes to the artifact -- so that there's an easy way for devs to look at the output of the build.

becw:
This work is included in the [2.0.0-alpha3 release](https://github.com/palantirnet/the-build/releases/tag/2.0.0-alpha3).

PR #115, opened 2018-11-14
Allow keeping the result of an artifact build without cleaning it up

becw:
Addresses #109.

I had to fix two bugs while working on this:
* The composer location stuff from #110 was failing because it needed the full path to composer (even when I didn't configure the `composer.composer` property)
* The `artifact-resetState` target was failing on some of the cleanup `<exec>` statements

Other than that, this changes the artifact build process so that instead of prompting the user with `Push artifact changes to the '${artifact.git.remote_branch}' branch? y,n`, it offers this prompt:
```
Push artifact changes to the 'artifact-test' branch?
  0: push
  1: keep
  2: discard
 [0]
```

This defaults to pushing if you hit enter.

* `push` pushes the artifact to the remote
* `keep` doesn't push, but leaves the generated artifact in place so that you can examine it
* `discard` reverts to the previous version of the artifact

### DIFF
--- a/docs/artifacts.md
+++ b/docs/artifacts.md
@@ -16,10 +16,10 @@ All artifact configuration should be in your project's base properties file, `.t
 
 ## Runtime flags
 
-* `push` - Value should be `y` or `n`. When this flag is provided, it will bypass the "Push artifact changes?" prompt.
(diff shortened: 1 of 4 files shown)


becw:
Merging this in order to get these bugfixes, and cut another alpha tag.

Issue #105, opened 2018-10-12
Provide drush aliases for multisites


becw:
Followup to #100.

Create Drush alias configuration when adding new multisites. See https://github.com/palantirnet/the-build/blob/c7a67f05ac4e8e38f8166f0bb1948a9baa54d8b8/tasks/install.xml#L224-L267, https://github.com/palantirnet/the-build/blob/c7a67f05ac4e8e38f8166f0bb1948a9baa54d8b8/tasks/drupal.xml#L350-L371

We should decide about whether to go to Drush 9 or not before working on this.

becw:
This work is included in the [2.0.0-alpha3 release](https://github.com/palantirnet/the-build/releases/tag/2.0.0-alpha3).

PR #116, opened 2018-11-14
Drush aliases for automatically created multisites

becw:
Addresses #105.

* Creates a drush aliases file in `drush/` when installing the-build
* Adds a new alias to the drush alias file when creating a new multisite
* Adds the multisite config to the-build's properties file automatically

To test:

1. Install the-build with `vendor/bin/the-build-installer`
2. Run `drush sa` -- it should list a `sitename.local` alias for your site
3. Create a new multisite with `phing drupal-add-multisite`
4. Run `drush sa` again -- there should be a new `sitename.local.NEWSITE` alias for your new multisite
5. Also, your `.the-build/build.default.properties.yml` file should have the Drupal sites configuration for your new multisite

A couple notes...
* This code got super messy because whitespace within XML is awkward

### DIFF
--- a/defaults/templates/projectname.aliases.drushrc.php
+++ b/defaults/templates/projectname.aliases.drushrc.php
@@ -8,4 +8,5 @@ $aliases['local'] = [
   'uri' => '${drupal.sites.default.uri}',
 ];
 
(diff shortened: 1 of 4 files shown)


becw:
Merging so that I can cut a new alpha tag.

Issue #114, opened 2018-11-13
"Building the build" documentation (development documentation)


froboy:
@becw I realized I'm not totally sure what the best way to work on the build is. Most of the time I think I've done some git acrobatics on an existing project, but if there was some documentation on how best to contribute to the project, that'd be awesome.

Potential topics:
- Getting a test project up and running with the build
- Working on the build in an existing project
- Potential pitfalls
- Common scenarios to test for

Thanks!

PR #118, opened 2018-11-20
Add info on how to run the-build for development.

becw:
Addresses #114.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -15,11 +15,11 @@ The major thing to watch out for here is that your copy of the-build is temporar
 
 ## Testing on an existing site
 
-You can clone the-build into the vendor directory of an existing site. If it doesn't have a version of the-build already installed, you'll need to require it with `composer require palantirnet/the-build:dev-release-2.0` first.
+You can clone the-build into the vendor directory of an existing site. If it doesn't have a version of the-build already installed, you'll need to require it with `composer require palantirnet/the-build:dev-release-2.0 --dev` first.
 
 ```
 cd my-site
-composer require palantirnet/the-build:dev-release-2.0 --prefer-source
+composer require palantirnet/the-build:dev-release-2.0 --prefer-source --dev
 ```
 
 At this point, `vendor/palantirnet/the-build` should be a clone of `git@github.com:palantirnet/the-build.git`.
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,0 +1,31 @@
+# Developing on the-build
+
+To develop and test changes to the-build, you'll generally need to have a Drupal codebase to run the-build on.
+
+1. Choose a site to test against
+2. Check out the-build as a git repository into the site's `vendor/` directory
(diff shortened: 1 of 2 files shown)


becw:
The work required to update from one version of the-build to the next totally depends on what the changes are between the versions. For the-vagrant, I include this info in the release notes, and I've started to do this in the-build as well. Basically:
* If the release includes changes to templates, you have to re-run the install script OR manually apply the changes. Sometimes these changes are optional, other times things will be broken until you reinstall -- and this is information that I include in the release notes.
* If the release only includes changes to library code, then you can just update the package to make the changes available to your project.

I think this info is kind of fiddly, which is why I include it with release notes -- it works better when it's applied to a specific situation.

Re: the change you made on how to check out the repository, I'm not sure about that -- I included the list as a short intro/overview to the doc. Running `composer require palantirnet/the-build:dev-release-2.0 --prefer-source --dev` will check the-build out as a repo, so you don't necessarily need to do that manually. Do you think we should include both methods of getting the repository into place?

becw:
And, re-reading the diff of this PR, I realize that the info is a little all over the place -- I totally see why you added that info, even if I think it should be in a different place :)

froboy:
@becw yea, rereading the instructions I see that the top where I added was meant to be a summary, then the details are below. I also think I missed the `--prefer-source` when I ran `composer require` the first time and was confused when `.git` was missing. I've removed those extra instructions and made sure the two composer commands match. That make more sense?

docs/development.md
```diff
@@ -0,0 +1,31 @@
...
+```
+cd my-site
+composer require palantirnet/the-build:dev-release-2.0 --prefer-source --dev
+```
+
+At this point, `vendor/palantirnet/the-build` should be a clone of `git@github.com:palantirnet/the-build.git`.
+
+## Creating a new site for testing
+
+Follow the [Quick Start guide for palantirnet/drupal-skeleton](https://github.com/palantirnet/drupal-skeleton#quick-start). Before installing the-build in step 3, update the required version of the-build:
```
jesconstantine: I ran into some composer dependency version collisions, namely: `the-build` requiring `drupal/coder:~8.2.12` and `drupal-skeleton` requiring `drupal/coder:^8.0`

Here are links to the `composer.json` package definition for each: 
* `the-build`: https://github.com/palantirnet/the-build/blob/release-1.0/composer.json#L16
* `drupal-skeleton`: https://github.com/palantirnet/drupal-skeleton/blob/drupal8/composer.json#L41

Since `the-build` is a dependency requirement of `drupal-skeleton` and requires `drupal/coder`, I'm curious if it's necessary to have `drupal/coder` also listed as a dev dependency in `drupal-skeleton`?

Once I manually did `composer require drupal/coder:~8.2.12` in my new site I was able to insntall the rest of the dependencies, including requiring the `release-2.0` branch from `the-build`

I did have some other issues continuing with the `drupal-skeleton` quick start steps, but I'll leave notes over in that repo.

PR #144, opened 2019-12-06
Adding file_temp_path to settings

shaal:
Per comment in https://www.drupal.org/project/drupal/releases/8.8.0, under "Other important update information".
```
The temporary file path is now a setting in settings.php, and is no longer stored in configuration. If your site is not using /tmp, you must update settings.php before running database updates:

$settings['file_temp_path'] = '/path/to/your/temporary/folder';
```

Original d.o issue: https://www.drupal.org/project/drupal/issues/3039026

### DIFF
--- a/targets/the-build.xml
+++ b/targets/the-build.xml
@@ -138,7 +138,7 @@
         <property name="drupal.site.database.database" value="${build.site}" />
         <property name="drupal.site.settings.file_public_path" value="sites/${drupal.site.dir}/files" />
         <property name="drupal.site.settings.file_private_path" value="../artifacts/private/${drupal.site.dir}" />
-        <property name="drupal.site.settings.file_temp_path" value="{$_ENV['TEMP']}" />
+        <property name="drupal.site.settings.file_temp_path" value="/tmp" />
         <property name="drupal.site.build.settings_dest" value="${drupal.root}/sites/${drupal.site.dir}/settings.build.php" />
         <property name="drupal.site.build.services_dest" value="${drupal.root}/sites/${drupal.site.dir}/services.build.yml" />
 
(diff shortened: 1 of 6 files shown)


targets/the-build.xml
```diff
@@ -138,6 +138,7 @@
         <property name="drupal.site.database.database" value="${build.site}" />
         <property name="drupal.site.settings.file_public_path" value="sites/${drupal.site.dir}/files" />
         <property name="drupal.site.settings.file_private_path" value="../artifacts/private/${drupal.site.dir}" />
+        <property name="drupal.site.settings.file_temp_path" value="{$_ENV['TEMP']}" />
```
becw: I think that this `{$_ENV['TEMP']}` environment variable is specific to Acquia ([docs](https://docs.acquia.com/acquia-cloud/develop/env-variable/)).

becw:
This is the right direction! We'll also need to cover the other config files.

targets/the-build.xml
shaal: @becw should we add a specific `tmp` for every vendor (acquia / pantheon)?
Is there a better way to do it?

becw: Each vendor does their own handling of this, so I think that we can customize this behavior in the `settings.*.php` templates, rather than providing it as a setting (because then everybody would have to figure out how it should be set for whatever provider they're using). The temp directory is generally an absolute path, rather than a relative path like the public and private files directories, so it tends to be a bit more host-specific.

For example, Pantheon already handles the temp directory as part of its scaffolding: https://pantheon.io/docs/tmp

For Platform.sh, the temp directory is specified in the `.platform.app.yaml` file (`/tmp`), so we can use that.

Even the `settings.build.php` should be fine to hard-code it, since projects can edit that template as necessary (it's copied to each project at `.the-build/drupal/settings.build.php`).

Issue #145, opened 2019-12-06
Remove config_installer usage


becw:
> The config_installer module is (partially) deprecated as Drupal core 8.6 now supports Installing Drupal from configuration.

See:

* https://www.drupal.org/node/2910643
* https://www.drupal.org/node/2897299

becw:
I'm going to put this on hold. The default install is on the `standard` profile, which doesn't work with `drush site-install --existing-config` -- it gives the following error:

> The selected profile has a hook_install() implementation and therefore can not be installed from configuration.

There's nothing we can really do here except continue using config installer for the time being -- oh well!

PR #147, opened 2020-01-21
Drush install from existing config

becw:
Use `drush site-install --existing-config` instead of `drush site-install config_installer`. See #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@
 
 * Added Drupal 8.8's `$settings['file_temp_path']` configuration to settings.*.php (#144)
 
+### Changed
+
+* `phing install` now uses `drush site-install --existing-config` instead of the [config_installer profile](https://www.drupal.org/project/config_installer) (#145)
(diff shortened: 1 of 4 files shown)


becw:
Oh, well, nevermind. `drush site-install --existing-config` gives the following error:

> The selected profile has a hook_install() implementation and therefore can not be installed from configuration.

Since the default/bare install is on the `standard` profile, there's nothing we can really do here except continue using config installer.

byrond:
@becw I think you mentioned we could try the `minimal` profile instead. It's nice to have some of the things that `standard` gives us, so I wonder how hard it would be to install with that, then switch to `minimal`. I just switched a project from `standard` to `config_installer` by using:
- `drush cset core.extension profile config_installer`
- `drush cr`
- `drush en config_installer`
- `drush cim -y` (the config in `core.extensions.yml` was already set to use `config_installer`)
I initially tried switching to `minimal` but was getting an error about missing block config for `admin_stark`.

PR #148, opened 2020-01-22
Use extension list instead of ignore for phpcs

byrond:
Due to https://www.drupal.org/project/coder/issues/3074176 ("Allow extensions to be overridden in contrib phpcs.xml file") included in Coder 8.37, the default list of extensions checked by `phpcs` has changed.

This removes the `ignore` configuration option and replaces it with `extensions` with a default that matches the original list in Coder.

**Note: This will require re-installing The Build or manually updating `build.xml` on projects.**

### Open questions

Should we add `js` to the list of extensions? The default behavior of Coder 8.3.7 checking JS files generated a lot of errors on one project where it was used, but it could be useful to have in the list.

Should we remove `ignore` from the defaults? This would cause projects that don't update their `build.xml` to start checking `.md` files, so it might be best to leave it in? Depending on the answer to the first question above, we could also add `*.js` to the list of ignored extensions to avoid checking files that weren't previously checked.

### DIFF
--- a/defaults.yml
+++ b/defaults.yml
@@ -233,7 +233,7 @@ phpcs:
   ignore: "*.md"
   
   # Comma-separated list of extensions to check in the PHP_CodeSniffer review.
-  extensions: "php,module,inc,install,test,profile,theme,css,info,txt,yml"
+  extensions: "php,module,inc,install,test,profile,theme,css,info,txt,yml,js"
 
 
 # Configuration for using PHP Mess Detector to check for general PHP best practices,
--- a/defaults.yml
+++ b/defaults.yml
@@ -228,9 +228,8 @@ phpcs:
   # Space-separated list of directories to review.
   directories: "${drupal.root}/modules/custom ${drupal.root}/themes/custom ${drupal.root}/profiles/custom"
 
-  # Comma-separated list of patterns for files and directories to exclude from the
-  # PHP_CodeSniffer review.
-  ignore: "*.md"
+  # Comma-separated list of extensions to check in the PHP_CodeSniffer review.
+  extensions: "php,module,inc,install,test,profile,theme,css,info,txt,yml"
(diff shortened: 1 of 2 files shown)


defaults.yml
```diff
@@ -228,9 +228,8 @@ phpcs:
   # Space-separated list of directories to review.
   directories: "${drupal.root}/modules/custom ${drupal.root}/themes/custom ${drupal.root}/profiles/custom"
 
-  # Comma-separated list of patterns for files and directories to exclude from the
```
becw: Existing build.xml files use this property -- so we should keep it, but add a comment saying it will be deprecated in 3.0 (which... we don't have a roadmap really so who knows when that would be...)

defaults/install/build.xml
```diff
@@ -132,7 +132,7 @@
         </phplint>
 
         <!-- Run PHP Code Sniffer. -->
-        <property name="phpcs.command" value="vendor/bin/phpcs --standard=${phpcs.standard} --ignore=${phpcs.ignore} ${phpcs.directories}" />
+        <property name="phpcs.command" value="vendor/bin/phpcs --standard=${phpcs.standard} --extensions=${phpcs.extensions} ${phpcs.directories}" />
```
becw: 👍

becw:
Re: your questions...

**Including js files in code reviews** - I think we should go ahead and add this. Existing projects will need to change their build.xml file to get this change anyway, and we should have this as a default on new projects

**Removing the `ignore` property from defaults** - We should keep this, again because of existing build.xml files, but add a deprecation warning.

byrond:
@becw I restored the default value for `phpcs.ignore`, added a deprecation warning and note about not using the option with Coder `>=8.3.7`, and added `js` to the default list of extensions that are checked.

PR #150, opened 2020-02-21
Allow using Drush 10 with the-build

becw:
Let's see how these tests go.


becw:
As it turns out, the tests don't test dependencies, since it relies on moving the current development into an existing site -- because you can't necessarily `composer require` a new development branch immediately (it takes 1-5 minutes for packagist to pick it up).

But, in my local testing this works fine with Drush 10.

PR #153, opened 2020-04-14
install drupal-check on circle

byrond:
- Removes the `drupal-check` installation step from the CircleCI config that tests The Build itself.
- Adds a step in the CircleCI config template to install `drupal-check`.
- Matches the installation method for `drupal-check` with The Vagrant. Installing via Composer can lead to issues as dependencies change. For example:
```
     [exec]  ------ --------------------------------------------------------------
     [exec]          Internal error: Call to undefined method
     [exec]          PhpParser\Comment\Doc::getEndLine()
     [exec]          Run PHPStan with --debug option and post the stack trace to:
     [exec]          https://github.com/phpstan/phpstan/issues/new
     [exec]  ------ --------------------------------------------------------------
```

See https://github.com/palantirnet/the-build/pull/133/files#diff-1d37e48f9ceff6d8030570cd36286a61 and https://github.com/palantirnet/drupal-skeleton/pull/109

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,19 +46,6 @@ jobs:
           name: Configure URL in /etc/hosts
           command: echo 127.0.0.1 ${CIRCLE_PROJECT_REPONAME}.local | sudo tee -a /etc/hosts
 
-      # Note: phing and drupal-check have mutually exclusive requirements.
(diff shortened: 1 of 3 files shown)


becw:
One more request: can you update https://github.com/palantirnet/the-build/blob/develop/defaults/install/the-build/build.circleci.yml#L17-L18 to match?

PR #154, opened 2020-05-18
Drupal check followup - fix the-build's CI tests

becw:


becw:
@byrond and I referenced the version of drupal-check pinned in the-vagrant (https://github.com/palantirnet/the-vagrant/pull/64/files) and also a thread in the drupal-check issue queue that suggests pinning to a specific version rather than using the newest releases (https://github.com/acquia/blt/pull/4001).

PR #160, opened 2020-07-08
Remove drupal-check; we're now recommending upgrade_status instead.

becw:
My main concern about this change is that we're not replacing the functionality directly with upgrade_status, since it seems like that's the kind of thing you only install when you need it.

Thoughts?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,20 +46,6 @@ jobs:
           name: Configure URL in /etc/hosts
           command: echo 127.0.0.1 ${CIRCLE_PROJECT_REPONAME}.local | sudo tee -a /etc/hosts
 
-      # Note: phing and drupal-check have mutually exclusive requirements.
-      # It'd be better to add drupal-check as a dependency of the drupal project
-      # rather than as part of the virtual environment, but this will have to do
-      # for now. Also note, drupal-check is added as part of the-vagrant so it
-      # is available to run within our VM.
-      # Note 2: drupal-check is pinned to a known stable version.
-      - run:
-          name: Install drupal-check
-          command: |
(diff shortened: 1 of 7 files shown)


byrond:
@becw I agree that Upgrade Status probably isn't something we would include in The Build, because it's only useful for that one purpose. Deprecation warnings are only one component of Drupal Check, which also runs PHPStan for code analysis and style checking. If we have all the other functionality covered by `phpcs` and `phpmd`, maybe we could remove Drupal Check, but deprecation warnings can be useful outside the major upgrade path.

(Most of that above was me thinking through things, but it seems to me Drupal Check would be fine to leave in The Build. Maybe we should just add some documentation on how to disable it, if desired?)

becw:
Hmm, you're right -- we shouldn't remove this if it's still useful for that kind of thing!

eric-schmidt:
@becw @byrond This is actually causing issues when spinning up new sites, as it is not added to ddev, so I would say we could remove this. We'll also need to remove a few lines from drupal-skeleton as well: https://github.com/palantirnet/drupal-skeleton/pull/120.

If we do not want to remove this, we will need to add this package via composer, and update references to it in a few files in the codebase, which I'm happy to do if need be! I've got a good example to reference, as I did this for Hampshire 👍

byrond:
@eric-schmidt @becw my vote is to try and update Drupal Check and install it via Composer. The version we were using (1.0.9) is really old, so maybe the conflicts with Phing are no longer present? 🤞

eric-schmidt:
@byrond I was able to get Drupal Check working just fine in this manner for Hampshire, so I'm happy to try and get those changes replicated here 👍

eric-schmidt:
Since we've decided to still include drupal-check (see https://github.com/palantirnet/the-build/pull/181), I"m going to go ahead and close this outdated issue out 👍

PR #161, opened 2020-07-29
Require the MySQL 5.6 driver for Acquia projects.

becw:
### Testing instructions

1. composer create-project with palantirnet/drupal-skeleton: `composer create-project palantirnet/drupal-skeleton test-the-build dev-develop --no-interaction`
2. cd into your new site: `cd test-the-build`
3. require this branch of the-build: `composer require --dev palantirnet/the-build:dev-mysql-5.6-driver-for-acquia`
4. run the installer for the-build: `vendor/bin/the-build-installer`
5. when it prompts you for the hosting platform, type `acquia`
6. (you'll see an error since the installer doesn't have access to mysql -- that's normal and expected!)
7. now look at your composer.json -- the `drupal/mysql56` package should be included

You would still need to enable the driver after installing drupal (`drush en mysql56`).

### DIFF
--- a/targets/install.xml
+++ b/targets/install.xml
@@ -262,6 +262,15 @@
         <switch value="${build.host}">
             <case value="acquia">
                 <copy file="${build.thebuild.dir}/defaults/install_site/drupal/settings.acquia.php" todir="${build.dir}/${drupal.root}/sites/${drupal.site.dir}" />
+                <!--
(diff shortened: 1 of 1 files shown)


becw:
This is no longer needed since as of 8/5/2020 Acquia is provisioning MySQL 5.7 for new sites. See https://docs.acquia.com/cloud-platform/create/install/drupal9/.

PR #163, opened 2020-11-10
fix composer cache path on circle

byrond:
The Composer cache path seems to have changed on CircleCI, most likely due to the update to Composer 2.
```
#!/bin/bash -eo pipefail
ls -1R ~/.composer/cache/ > /tmp/composer-cache.txt
ls: cannot access '/home/circleci/.composer/cache/': No such file or directory
Exited with code exit status 2
CircleCI received exit code 2
```

Updating the Composer cache path in the CircleCI `config.yml` fixes the issue.


becw (approved):
👍

PR #174, opened 2021-11-12
Automatically run drush inside of ddev

becw:
If this is a ddev project, and we're outside of ddev, run ddev's drush command from phing instead of running drush directly.

After merging, we'll need to update and merge https://github.com/palantirnet/drupal-skeleton/pull/121 in palantirnet/drupal-skeleton.

This actually seems to work! You can test this with:

### Set up the codebase
```
cd example
composer require --dev palantirnet/the-build:dev-detect-ddev
ddev start
```

### Install Drupal at the prompt
When you're prompted to _Install Drupal now (y,n)?_, say _y_.

### The site has been installed without ever logging in to ddev
Visit https://example.ddev.site/

### Run other phing commands
For example:
```
vendor/bin/phing install -Ddrupal.validate_clean_config.bypass=yes
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "extra": {
         "patches": {
             "phing/phing": {
-                "Support relative symliks in Phing": "patches/phing-relative-symlinks.patch"
+                "Support relative symliks in Phing": "https://raw.githubusercontent.com/palantirnet/the-build/7cdc28b6019fb88a0604261366f9ea35f1e21d96/patches/phing-relative-symlinks.patch"
             }
         }
     }
--- a/composer.json
+++ b/composer.json
@@ -12,9 +12,10 @@
         "bin/the-build-installer"
     ],
     "require": {
-        "palantirnet/phing-drush-task": "^1.1",
+        "cweagans/composer-patches": "^1.7",
         "drupal/coder": "^8.3.6",
         "drush/drush": "^9 || ^10",
+        "palantirnet/phing-drush-task": "^1.1",
         "pear/http_request2": "^2.3",
         "pear/versioncontrol_git": "@dev",
         "phing/phing": "^2.14",
@@ -27,5 +28,12 @@
     },
     "config": {
         "sort-packages": true
+    },
+    "extra": {
+        "patches": {
+            "phing/phing": {
+                "Support relative symliks in Phing": "patches/phing-relative-symlinks.patch"
+            }
+        }
     }
 }
--- a/defaults/install/build.xml
+++ b/defaults/install/build.xml
@@ -45,7 +45,7 @@
 
         <!-- Include styleguide resources in the theme. This approach will symlink
              resources in development environments, and copy them for artifact builds. -->
-        <!-- <includeresource source="styleguide/source/assets/css" dest="${drupal.root}/themes/custom/example_theme/css" /> -->
+        <!-- <includeresource relative="true" source="${build.dir}/styleguide/source/assets/css" dest="${build.dir}/${drupal.root}/themes/custom/example_theme/css" /> -->
     </target>
 
 
--- a/patches/README.md
+++ b/patches/README.md
@@ -1,0 +1,9 @@
+### phing-relative-symlinks.patch
+
+Source: [https://github.com/phingofficial/phing/pull/695](https://github.com/phingofficial/phing/pull/695)
+
+This patch updates Phing's SymlinkTask to allow creating relative symlinks, using the syntax:
+
+```
+<symlink link="path/to/destination" target="path/to/target" relative="true" />
+```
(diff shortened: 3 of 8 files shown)


eric-schmidt:
@becw I think there might be a mismatch with the DDEV server name. When trying to spin this up, DDEV sets its URL to http://drupal-skeleton.ddev.site; however, the install commands seem to be trying to access http://example.ddev.site (or whatever you named the project during the `composer create-project` step) -- could that cause issues?

I'm also seeing errors when trying to run `ddev drush sql-query`, which is weird because `ddev drush status` works just fine 🤔 I'm wondering if some of the issues people were running into with `drush ev` in [this issue](https://github.com/drud/ddev/issues/1648#issuecomment-513389861) are related...

becw:
That drush ev issue looks likely! The underlying issue is also described here: https://github.com/drud/ddev/issues/2547

becw:
As far as the URL mismatch goes, that happens because the `.ddev/config.yaml` file (which is part of `palantirnet/drupal-skeleton`) has the URL hardcoded to `http://drupal-skeleton.ddev.site`, and the-build doesn't rewrite it when it is installed. So it should probably do that now, but that would happen separately from this issue I think.

eric-schmidt:
@becw Unfortunately, I'm running into some issues with the Devel version and it's compatibility with D9, and it's preventing the install. I actually ran into this the other day when I was setting up Hampshire and opened a PR for [the-build](https://github.com/palantirnet/the-build/pull/176) and [drupal-skeleton](https://github.com/palantirnet/drupal-skeleton/pull/118). Can you take a look at those and see if these are valid solutions?

eric-schmidt:
@becw I merged `develop` back into this branch in order to take advantage of the `devel` updates; however, I'm now met with this error when proceeding through the install step above 🤔 

Screen Shot 2021-12-02 at 10 18 05 AM

byrond:
@becw I tested this with an existing project, and it worked great for `phing install`, running `drush site-install` inside DDEV and everything else outside.

@eric-schmidt I haven't tested this on a first Drupal install yet, but I think this is an issue with the symlink that is created outside of DDEV here:
https://github.com/palantirnet/the-build/blob/develop/targets/drupal.xml#L258

I'm not exactly sure how we can figure out what the actual directory name will be inside of DDEV from outside. Maybe we could make it relative somehow?

byrond:
Does this matter (maybe this is exactly what you were already discussing above)?
```
    [drush] Executing 'ddev drush --uri="https://example.ddev.site" sql-query "\\"SELECT 1\\""'...
bash: -c: line 2: syntax error: unexpected end of file
                                                      Failed to run drush --uri=https://example.ddev.site sql-query \SELECT 1\: exit status 2
```

It seems to correctly detect whether Drupal has been installed in spite of the error.

I can confirm that the symlink is created from the host machine perspective and results in:
`the_build_utility -> /Users/byron/example/vendor/palantirnet/the-build/defaults/standard/modules/the_build_utility`

Here is the step from earlier in the Phing task:
```
  [symlink] Linking: /Users/byron/example/vendor/palantirnet/the-build/defaults/standard/modules/the_build_utility to /Users/byron/example/docroot/modules/contrib/the_build_utility
```

I'm looking into a way to possibly fix that symlink.

byrond:
Blergh. It looks like relative symlinking is only available in Phing 3:
https://github.com/phingofficial/phing/commit/1adb2adf9e6b30e9543259a34a1e6a8737ed146a

That PR fixes this issue, which is exactly what we are trying to do, though:
https://github.com/phingofficial/phing/issues/684

byrond:
So, it turns out we can use the changes in the commit above as a patch for Phing 2, and it solves this issue with the missing `the_build_utility` module. Now, I'm just not sure the best way to apply it, since `cweagans/composer-patches` would need to go in the `composer.json` of the skeleton along with the `"patches": { }` section. However, it's really a requirement for `the-build`.

byrond:
@becw @eric-schmidt, I created a PR (https://github.com/palantirnet/the-build/pull/179) to patch Phing, but it will also require updating the skeleton to add `"enable-patching": true` to its `composer.json`, so the patch included in this module's `composer.json` will be installed.

I had to use a URL instead of a relative path, because Composer will install the patch from the perspective of the project root, not the-build's root.

It would be better if the patch could be added to Phing 2 in the vendor's repo, but I'm not sure how long that would take, and we would need to wait on a release to use it in the-build.

byrond:
I confirmed the patch gets installed when running:
`composer create-project palantirnet/drupal-skeleton example dev-composer-patches-dependencies --no-interaction`

After that, all that is needed to test is `ddev start` and `vendor/bin/the-build-installer`.

I'm opening a PR against the skeleton for testing.

byrond:
I opened a PR for Phing against what I think is the branch for version 2. I don't know if they are still doing bugfixes and releases, so it may not get merged.
https://github.com/phingofficial/phing/pull/1654

byrond:
@becw @eric-schmidt we will need to do something similar for Behat when running `phing test`.

PR #177, opened 2021-11-29
Use new CircleCI convenience image for PHP

byrond:
See https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

Todo:

- [x] Update `defaults/install/.circleci/config.yml`
- [x] Maybe add a comment with a link to the orb documentation for installing browsers?
- [x] ~Maybe add a comment about when the mysql image will get a cimg version?~ Update for new MySQL image.
- [x] The devel version issue should be resolved in the `develop` branch now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
       - run:
           name: Install packages
           command: |
-            sudo sudo apt-get update && apt-get install -y libpng-dev default-mysql-client
+            sudo apt-get update && apt-get install -y libpng-dev default-mysql-client
       - run:
           name: Configure PHP
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     working_directory: ~/example
     docker:
-      - image: circleci/php:7.4-node-browsers
+      - image: cimg/php:7.4-browsers
       - image: circleci/mysql:5.7-ram
         command: --max_allowed_packet=16M
         environment:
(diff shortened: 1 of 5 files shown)


becw:
I added some todos to the description but this change sounds good!

byrond:
@becw, even though https://github.com/palantirnet/the-build/pull/176 is merged, Composer grabs the latest release of The Build (which is 4.0-alpha4) during install. That version doesn't have the Devel fix, so https://github.com/palantirnet/drupal-skeleton/pull/119 will fail until there is a new release of 4.x of The Build.

[Edited to correct the installed version in the skeleton and clarify that the release doesn't need to be stable.]

byrond:
@becw, I created https://github.com/palantirnet/the-build/releases/tag/4.0-alpha5 and https://github.com/palantirnet/drupal-skeleton/pull/119 is passing now without specifying the dev branch.

.circleci/config.yml
```diff
@@ -24,14 +24,10 @@ jobs:
           command: |
             sudo apt-get update
             sudo apt-get install -y libpng-dev default-mysql-client
```
byrond: This probably needs a `sudo apt-get update && ` to match https://github.com/palantirnet/drupal-skeleton/pull/119/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R25

byrond: Well, it already has one right above, but I moved them to the same line to stay consistent with other config.yml files.

byrond:
@becw I just made one last update to the example deployment config snippets. Otherwise, I think this is ready to go.

becw (approved):

PR #178, opened 2021-12-01
Remove profiles/custom/ directory from code checks

becw:
Remove profiles/custom/ directory from code checks, because they break when it's not present and we don't generally use this directory.


eric-schmidt:
Looks good to me!

PR #179, opened 2021-12-23
Patch Phing to fix relative symlinks

byrond:
#### Description

* Fixes issues discovered in https://github.com/palantirnet/the-build/pull/174.
* Patches Phing to fix support for relative symlinks (according to https://github.com/phingofficial/phing/pull/695, which was only added to Phing 3).
* Uses a relative symlink for `the_build_utility` module, so it will be correct inside the DDEV container when created outside.
* Updates `IncludeResourceTask` to use `SymlinkTask` in order to take advantage of required relative symlinks.

This also requires adding `cweagans/composer-patches` and the following line to `composer.json`, so the skeleton will need to be updated:
```
"extra": {
    "enable-patching": true
}
```

#### Testing instructions

* Create a new project using Composer, or use an existing one. Backup your database if using an existing project, and you want to keep it.
* `composer require cweagans/composer-patches`
* Add `"enable-patching": true` to the `"extra": { }` section of `composer.json`
* Change the version of The Build in `composer.json` to `"palantirnet/the-build": "dev-detect-ddev-byron"`.
* `composer update palantirnet/the-build` (You should see Composer remove Phing, reinstall it, and add the patch.)
* `ddev start` (if the project isn't already running)
* `ddev drush sql-drop`
* `vendor/bin/the-build-installer`
* Use defaults (or settings that match your existing project). Choose to install Drupal when prompted.

### DIFF
--- a/patches/phing-relative-symlinks.patch
+++ b/patches/phing-relative-symlinks.patch
@@ -1,0 +1,76 @@
+diff --git a/classes/phing/tasks/ext/SymlinkTask.php b/classes/phing/tasks/ext/SymlinkTask.php
+index f132b4f747..87844a0d09 100644
+--- a/classes/phing/tasks/ext/SymlinkTask.php
++++ b/classes/phing/tasks/ext/SymlinkTask.php
+@@ -206,6 +206,46 @@ public function isRelative()
+         return $this->relative;
+     }
+ 
++    /**
++     * Given an existing path, convert it to a path relative to a given starting path.
++     *
++     * @param string $endPath   Absolute path of target
++     * @param string $startPath Absolute path where traversal begins
++     *
++     * @return string Path of target relative to starting path
++     */
++    public function makePathRelative($endPath, $startPath)
++    {
++        // Normalize separators on Windows
++        if ('\\' === DIRECTORY_SEPARATOR) {
++            $endPath = str_replace('\\', '/', $endPath);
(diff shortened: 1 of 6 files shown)


src/TheBuild/IncludeResourceTask.php
```diff
@@ -32,6 +32,12 @@ class IncludeResourceTask extends \Task {
    */
   protected $dest = NULL;
 
+  /**
+   * Whether to create relative symlinks
+   *
+   * @var boolean
+   */
+  private $relative = true;
```
becw: This should be `protected` instead of private I think?

patches/phing-relative-symlinks.patch
```diff
@@ -0,0 +1,76 @@
+diff --git a/classes/phing/tasks/ext/SymlinkTask.php b/classes/phing/tasks/ext/SymlinkTask.php
```
becw, line 1: Can we add a README to the patches directory that explains where this patch comes from?

becw:
I had two comments/change requests but otherwise this looks good.

src/TheBuild/IncludeResourceTask.php
becw: I went ahead and made this change.

patches/phing-relative-symlinks.patch
becw: I went ahead and added a README.

PR #180, opened 2022-01-07
Change the behat command to use depending on whether ddev is available.

becw:
Followup to #174; this does the same thing for behat as we did for drush, so that behat is run in an environment that can access the Drupal application directly.

To test...

### Set up the codebase
```
cd example
composer require --dev palantirnet/the-build:dev-run-behat-inside-ddev
ddev start
```

### Install Drupal at the prompt
When you're prompted to _Install Drupal now (y,n)?_, say _y_.

The site should be available at https://example.ddev.site/

### Run the behat test command:
```
vendor/bin/phing behat
```

### DIFF
--- a/defaults/install/build.xml
+++ b/defaults/install/build.xml
@@ -103,7 +103,7 @@
 
     <!-- Target: behat -->
     <target name="behat" description="Run the Behat tests.">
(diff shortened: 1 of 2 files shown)


eric-schmidt (approved):
Worked like a charm for me!

byrond (approved):
Works for me as well!

PR #181, opened 2022-01-07
Add drupal check dependency

becw:
Add mglaman/drupal-check as a dependency, since Phing dependencies no longer conflict with drupal-check dependencies. This means that we no longer have to do gymnastics around finding the location of the drupal-check bin on whatever system we're running on (localhost, ddev, vagrant, circleci).

To test:

```
cd example
composer require --dev palantirnet/the-build:dev-add-drupal-check-dependency
vendor/bin/phing code-review
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "cweagans/composer-patches": "^1.7",
         "drupal/coder": "^8.3.6",
         "drush/drush": "^9 || ^10",
+        "mglaman/drupal-check": "^1.2",
         "palantirnet/phing-drush-task": "^1.1",
         "pear/http_request2": "^2.3",
         "pear/versioncontrol_git": "@dev",
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         "pear/http_request2": "^2.3",
         "pear/versioncontrol_git": "@dev",
         "phing/phing": "^2.14",
-        "phpmd/phpmd": "^2.4"
+        "phpmd/phpmd": "^2.4",
+        "phpspec/prophecy-phpunit": "^2"
     },
     "autoload": {
         "psr-0": {
--- a/defaults/install/the-build/build.circleci.yml
+++ b/defaults/install/the-build/build.circleci.yml
@@ -13,6 +13,3 @@ drupal:
 
 behat:
   args: "--profile=circleci --suite=default --strict --format=junit --out=/tmp/artifacts --tags=~@skipci"
-
-drupal-check:
-  bin: "/home/circleci/bin/drupal-check"
--- a/defaults/install/the-build/build.yml
+++ b/defaults/install/the-build/build.yml
@@ -53,13 +53,6 @@ drupal:
 behat:
   args: "--suite=default --strict"
 
-# Configuration for checking the site with the Drupal Checker code linter.
-drupal-check:
-  # Location of the drupal-check script. This script is currently installed globally
-  # on the-vagrant because of repository-level composer conflicts.
-  bin: "~/bin/drupal-check"
-
-
 # To build an artifact from your code, add the URL to your artifact git repository.
 # @see https://github.com/palantirnet/the-build/blob/release-2.0/defaults.yml
 #
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,10 +58,7 @@ jobs:
       # Use this copy of the-build
       - run:
           name: Replace the default version of the-build with this one
-          command: rm -r vendor/palantirnet/the-build
-
-      - checkout:
-          path: ~/example/vendor/palantirnet/the-build
+          command: composer require --dev palantirnet/the-build:dev-${CIRCLE_BRANCH}
 
       # Source cache - update when branch changes
       - save_cache:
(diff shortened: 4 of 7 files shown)


defaults/install/the-build/build.circleci.yml
```diff
@@ -13,6 +13,3 @@ drupal:
 
 behat:
   args: "--profile=circleci --suite=default --strict --format=junit --out=/tmp/artifacts --tags=~@skipci"
-
```
eric-schmidt, line 16: @becw On Hampshire I actually just updated this line to point to the vendor directory -- how come we wouldn't want this to still run during the Circle CI tests?

defaults/install/the-build/build.yml
```diff
@@ -53,13 +53,6 @@ drupal:
 behat:
   args: "--suite=default --strict"
 
-# Configuration for checking the site with the Drupal Checker code linter.
```
eric-schmidt, line 56: @becw Same goes for this -- would we not want this to run when building? I'm probably just misunderstanding :P

defaults/install/the-build/build.circleci.yml
becw: It would still run -- but `vendor/bin/drupal-check` is now the default value.

defaults/install/the-build/build.yml
becw: Again, it does still run -- but we don't need to set an environment-specific value with config anymore because we know it's going to be at `vendor/bin/drupal-check`.

eric-schmidt (commented):
@becw Thanks for those updated notes -- that all makes sense! Unfortunately, when I run through the testing instructions, I'm seeing the following errors:

Screen Shot 2022-01-21 at 2 38 02 PM

That seems to relate to this command: `<exec command="${drupal-check.command}" logoutput="true" checkreturn="true" />`, but it also seems to relate to some bad formatting that `drupal-check` itself doesn't expect. Any ideas?

composer.json
```diff
@@ -15,6 +15,7 @@
         "cweagans/composer-patches": "^1.7",
         "drupal/coder": "^8.3.6",
         "drush/drush": "^9 || ^10",
+        "mglaman/drupal-check": "^1.2",
```
byrond, line 18: Should this go in `require-dev`?

byrond:
@becw I'm trying this on the project I am currently resourced to and will report back. I am going to try with `require-dev` since we'll only use it on CircleCI and local development.

composer.json
byrond: Nope! Because this is The Build's `composer.json`, not a project's!

byrond:
This is working on my current project. There, I updated the CircleCI config, build config, and `build.xml`. I added drupal-check to the project's dev dependencies with Composer, and it is working fine there. Once this PR is merged, and we have a new release, I can update it in the project, then remove the explicit dependency from the project, as well as the lines from the build config.

byrond:
@becw I also ran through the testing steps for a new project and didn't have the same error as @eric-schmidt.

```
example > drupal-check:

     [echo] $> vendor/bin/drupal-check docroot/modules/custom/
    0 [▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░]
     [exec] 
     [exec] 
     [exec]  [OK] No errors
     [exec] 

example > drupal-check:

     [echo] $> vendor/bin/drupal-check docroot/themes/custom/
    0 [▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░]
     [exec] 
     [exec] 
     [exec]  [OK] No errors
     [exec] 

BUILD FINISHED

Total time: 5.4969 seconds
```

byrond (approved):

eric-schmidt:
This appears to be an issue with drupal-check itself, as I get the same error when manually running it outside of phing. What is weird is that I can run the exact same command using the exact same version of drupal-check on other sites w/o issue 🤔 

Screen Shot 2022-01-28 at 1 29 12 PM

byrond:
@eric-schmidt what version of PHP are you using on your host?

byrond:
@eric-schmidt @becw I added this to an existing project, and although I had to change my `memory_limit` in `php.ini`, it worked as expected. I tried again on a new example project, and it is still working for me there as well.

byrond:
When using this on an existing project, I encountered the following error:
```
     [exec]      Child process error (exit code 1): Drupal requires Prophecy PhpUnit when
     [exec]      using PHPUnit 9 or greater. Please use 'composer require --dev
     [exec]      phpspec/prophecy-phpunit:^2' to ensure that it is present.
```
Once I added the dependency, `drupal-check` worked as expected.

byrond:
@becw I added `phpspec/prophecy-phpunit:^2` as a dependency of the-build, since that message has come up a few times while using this branch.

byrond:
The skeleton now installs version 3.0.4 of `composer/xdebug-handler`, which conflicts with `drupal-check` dependencies. I ran into this with a project using this branch, so I reran tests here. They failed, as expected. I will look for an update to `drupal-check`.

byrond:
`composer/composer` requires either version 2 or 3 of `composer/xdebug-handler`, so version 3 gets installed during `composer create-project`. The `composer require` step needs to be able to downgrade that to version 2, since `drupal-check` doesn't support 3 yet. So, I added `--with-all-dependencies` to the `composer require` step in the CircleCI build job to allow it to happen.

@becw @eric-schmidt are we ready to merge this yet?

PR #183, opened 2022-02-02
Updates phpcs config in defaults.yml

### LINKED ISSUE #157: --ignore flag for phpcs is documented in defaults.yml but not used by default in build.xml
defaults.yml documents providing an ignore list for phpcs, but that variable is not used in the default build.xml.

byrond:
* Fixes https://github.com/palantirnet/the-build/issues/157 by updating the comments to reference the new `extensions` config.
* Removes the deprecated `ignore` config.

### DIFF
--- a/defaults.yml
+++ b/defaults.yml
@@ -227,13 +227,17 @@ phplint:
 # @see https://www.drupal.org/project/coder
 #
 # These values are used in the defaults/build.xml template:
-#   $> phpcs --standard=${phpcs.standard} --extensions=${phpcs.extensions} ${phpcs.directories}
+#   $> phpcs --standard=${phpcs.standard} --extensions=${phpcs.extensions} --ignore=${phpcs.ignore} ${phpcs.directories}
 phpcs:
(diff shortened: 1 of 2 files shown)


byrond:
Fooey! I actually found that I needed the `--ignore` option for a module that will include a `vendor` and `node_modules` directory. I'm going to rework this to add the option back in and use it in `build.xml`.

byrond:
Never mind! I can just move my code-review step earlier in the build, so there are no vendor and node_modules directories to be scanned.

byrond:
Argh. I think I'm switching sides again and saying we could use this. Moving the code-review step works for CircleCI, but during development, if your module has Composer or Node dependencies, they will need to be ignored.

becw:
Another option here would be to provide a default `phpcs.xml` file, and remove all of the customized phpcs config and commands from the-build. That way, the `build.xml` file could run phpcs without arguments, developers could easily run the same phpcs config by running phpcs directly, and it would be more obvious that phpcs was set up for a project...

becw:
I merged #185 instead.

PR #184, opened 2022-03-11
Fix shell script installer

becw:
Starting in Composer 2.2, running `vendor/bin/the-build-installer` resulted in the following error, even when running the installer from the correct location:

> usage: dirname string [...]
> Please run this command from your project root.

This seems to be because Composer 2.2.0 included changes to the way scripts in the `vendor/bin/` directory are proxied/run, which broke the installer wrapper script. I'm not sure if there's a better way to fix this -- the updated script still assumes that scripts are located in the `vendor/bin/` directory.

* Script proxying was added in Composer 2.2.0 -- previously, scripts were symlinked into the `vendor/bin/` directory
* The environment variable `COMPOSER_BIN_DIR` was added in Composer 2.2.2
* The environment variable was changed to `COMPOSER_RUNTIME_BIN_DIR` in Composer 2.2.7

### Testing instructions

```
cd example

# Pause here and update .ddev/config.yaml to change the "name" from "drupal-skeleton" to "example"

composer require palantirnet/the-build:dev-fix-installer-dirname-error

# Follow the prompts to set up the site and install Drupal.
```

--> The project `example` should be set up, and ddev should be running with the installed site.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,6 +28,10 @@ jobs:
           command: |
             echo "sendmail_path=/bin/true" | sudo tee -a "/etc/php.d/circleci.ini"
       - run:
+          name: Update Composer
+          command: |
+            sudo composer self-update
+      - run:
           name: Create artifacts directory
           command: mkdir /tmp/artifacts
       - run:
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ _Note: If you are starting a new a project, you may be looking for the [drupal-s
 $> composer require palantirnet/the-build
 ```
 
+Composer 2.2.2 or greater is required for the-build.
+
 ## Setting up
 
 Install the default templates and phing targets to your project:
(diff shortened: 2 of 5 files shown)


becw:
The way I had to write this script makes me feel like I'm doing something wrong. But, the script can't reliably find `phing` without being able to reliably find either the `vendor/bin/` directory or itself, and between Composer 2.2.0 and previous versions, the script location changed because Composer started wrapping/proxying scripts, and then an env variable was added in 2.2.2, and then the env variable was changed in 2.2.7...... so I'm not sure what else I could have done here.

eric-schmidt (approved):
@becw This is all working for me! I did also test downgrading Composer to 2.1.0 just to see what would happen now that we're requiring ^2.2.2. It did cause the `composer require palantirnet/the-build:dev-fix-installer-dirname-error` command to fail out as expected, but I would be curious how it fails once this is merged in and is a part of the `composer create-project palantirnet/drupal-skeleton` command -- so we may want to test that once this is rolled out, just to see how it communicates the issue.

PR #185, opened 2022-03-11
Move phpcs configuration from the-build into a phpcs.xml file

becw:
Alternative to #183.

This pulls out all of the phpcs config from the-build and puts it in a default `phpcs.xml` file. This should allow running the project's phpcs config by just running `phpcs` (or `vendor/bin/phpcs`) without any special arguments. It also makes it more visible to developers new to a project that phpcs is available.

This config file should probably be tested in an actual project.

If a project updates from an older version of the-build to a version that includes these changes, they would need to:
* Update the `build.xml` to use the new command
* Copy the new phpcs.xml file into the project

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,11 @@
         }
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "cweagans/composer-patches": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true
+        }
     },
     "extra": {
         "patches": {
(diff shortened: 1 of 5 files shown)


becw:
Ooh, maybe we can do this for phpmd too.

eric-schmidt:
@becw I tried testing this on a live project; however, I didn't have a good one where I could actually push something up, so I was just testing by running the `phpcs` command in the repo root; however, it errors out on a ton of stuff that `phpcs` wouldn't typically notice. Running the command in the repo root should pick up this XML file in the same way that the build step would, is that right?

Maybe it would be best to just find a project where I can make this change in a PR and push it up to truly test the build tooling 🤔

becw:
I think Eric was getting more coding errors/suggestions because the old coding standard was just "Drupal", and this phpcs file has both "Drupal" and "DrupalPractice". I think for new projects, having both in place is better.

PR #190, opened 2022-07-27
Fix the installer when using "other"

byrond:
`drupal.root` is not set when selecting the "other" option during the-build installation. This results in the following error:

```
install > setup-web-root:

     [echo] Moving the docroot/ directory to ${drupal.root}/
sh: 1: Bad substitution
[phingcall] /var/www/html/vendor/palantirnet/the-build/targets/install.xml:137:12: /var/www/html/vendor/palantirnet/the-build/targets/install.xml:145:125: Task exited with code 2

BUILD FAILED
/var/www/html/vendor/bin/../palantirnet/the-build/targets/install.xml:86:44: Execution of the target buildfile failed. Aborting.
```

* Defines a default value for `drupal.root` during installation. The `switch` structure could be simplified, but this method leaves room for other properties to be set for each environments if needed.
* Modifies the `setup-web-root` target to check for a `core` subdirectory, because DDEV will create whatever it is configured to use as its `docroot` if it doesn't already exist, and it will add `settings.ddev.php` there. Also, we remove the new docroot before moving the old one.

### Testing

- `composer create-project palantirnet/drupal-skeleton example dev-develop --no-interaction`
- `cd example`
- `composer require --dev palantirnet/the-build:dev-fix-installer-for-other`
- Follow the instructions in https://github.com/palantirnet/drupal-skeleton#steps (starting at step 2).
- Choose "other" during the-build-installer.

### DIFF
--- a/targets/install.xml
+++ b/targets/install.xml
@@ -144,7 +144,7 @@
                 <available file="${build.dir}/docroot/core" type="dir" property="drupal.root.old" value="docroot" />
 
                 <echo msg="Moving the ${drupal.root.old}/ directory to ${drupal.root}/" />
-                <exec command="rm -rf ${drupal.root}" dir="${build.dir}" checkreturn="true" logoutput="true" />
+                <exec command="rm -rf ${drupal.root}" dir="${build.dir}" />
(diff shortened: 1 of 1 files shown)


becw:
I made one minor change. Other than that, 👍

PR #191, opened 2022-08-01
test the-build with drupal-10

byrond:


agentrickard:
@byrond This failed on Circle -- I also don't know that we can commit it to `develop` as-is, since it replaces Drupal 9. 

Is the idea that this will become the default after D10 is released?

byrond:
Closing in favor of #201

PR #192, opened 2022-10-03
Support drush 11

agentrickard:
Adds optional support for Drush 11.

See https://www.drush.org/latest/install/#drupal-compatibility


becw:
We'll also need to update the drush requirement in palantirnet/phing-drush-task -- Byron's Drupal 10 PR there has that.

PR #197, opened 2022-11-10
Update coding standards in the-build.

becw:
Let's see if this works.


byrond (commented):
@becw this looks good, and I tested the installer, the build, and the install tasks. ~It looks like we're losing some documentation found in some of the task file comment blocks. Is that ok? Do we have those things documented elsewhere?~

Oh, never mind! I found it here: https://github.com/palantirnet/the-build/blob/0217134c57c29b09d617c3938221263cb16c9017/docs/tasks.md

So, it's good to remove from code, so we don't have to maintain both sets!

PR #199, opened 2022-11-11
explicitly disable Pantheon build step

byrond:
Applications created on Pantheon now use an integrated Composer build: https://pantheon.io/docs/guides/integrated-composer

Since we push complete artifacts, we need to disable this build step to ensure it doesn't attempt to run against our artifact. This updates the default `pantheon.yml` that gets installed with the-build to disable the build step.


becw (approved):
👍

PR #210, opened 2023-01-13
Tech 35 pantheon install

agentrickard:
### Testing steps

* `composer create-project palantirnet/drupal-skeleton example dev-develop --no-interaction`
* `cd example`
* `composer require --dev "palantirnet/the-build:dev-TECH-35-pantheon-install"`
* Edit `.ddev/config.yaml` and change `name` to `tech35`
* Uncomment the lines for Pantheon, and comment out the lines for Acquia (containing `docroot`, `webserver_type`, and `mariadb_version`/`mysql_version`).
* `ddev start`
* `vendor/bin/the-build-installer`
* Use `tech35` as the project machine name.
* Use `https://tech35.ddev.site` as the Drupal URI.
* Choose `pantheon` as the target platform.
* Choose to install Drupal when prompted.
* Verify the installer completes without errors.
* You should see the following during the first install when the Drupal root is moved:
```
install > setup-web-root:

     [echo] Copying settings.ddev.php to old drupal root.
```
* Verify Drupal is accessible at https://tech35.ddev.site
* If you're able, follow the steps above to switch back to Acquia.
* Note the `settings.ddev.php` is not copied this time, because it already exists in the old Drupal root.

### Remaining tasks

- [x] Revert the change to `.circleci/config.yml`
- [x] Temporarily remove Config Ignore from `drupal-first-install` until there is a stable release that supports Drupal 10: https://github.com/palantirnet/the-build/blob/TECH-35-pantheon-install/targets/drupal.xml#L278

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
       # Create a new project using the drupal-skeleton project
       - run:
           name: Create a new Drupal project
-          command: composer create-project palantirnet/drupal-skeleton example dev-minimum-stability-dev --no-interaction
+          command: composer create-project palantirnet/drupal-skeleton example dev-develop --no-interaction
           working_directory: ~/
 
       # Use this copy of the-build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
       # Create a new project using the drupal-skeleton project
       - run:
           name: Create a new Drupal project
-          command: composer create-project palantirnet/drupal-skeleton example dev-develop --no-interaction
+          command: composer create-project palantirnet/drupal-skeleton example dev-minimum-stability-dev --no-interaction
           working_directory: ~/
 
       # Use this copy of the-build
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "drush/drush": ">=9",
         "mglaman/drupal-check": "^1.2",
         "palantirnet/phing-drush-task": "^1.1",
-        "pear/http_request2": "^2.5",
+        "pear/http_request2": "^2.3",
         "pear/versioncontrol_git": "@dev",
         "phing/phing": "^2.14",
         "phpmd/phpmd": "^2.4",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "drush/drush": ">=9",
         "mglaman/drupal-check": "^1.2",
         "palantirnet/phing-drush-task": "^1.1",
-        "pear/http_request2": "^2.3",
+        "pear/http_request2": "^2.5",
         "pear/versioncontrol_git": "@dev",
         "phing/phing": "^2.14",
         "phpmd/phpmd": "^2.4",
--- a/targets/drupal.xml
+++ b/targets/drupal.xml
@@ -275,7 +275,6 @@ Or, you can specify the export file directly:
         <composer command="require" composer="${composer.composer}">
             <arg value="drupal/admin_toolbar" />
             <arg value="drupal/config_split" />
-            <arg value="drupal/config_ignore" />
             <arg value="drupal/devel" />
             <arg value="drupal/workbench" />
             <arg value="drupal/workbench_tabs" />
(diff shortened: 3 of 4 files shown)


agentrickard:
I don't know why the test is failing -- perhaps composer cache on CircleCI?

This fixes the Pantheon install problem, but I still get this (non-fatal) error:

```
[PHP Error] include_once(HTTP/Request2.php): Failed to open stream: No such file or directory [line 92 of /Users/rickard/Sites/example/vendor/phing/phing/classes/phing/tasks/ext/HttpTask.php]
[PHP Error] include_once(): Failed opening 'HTTP/Request2.php' for inclusion (include_path='/Users/rickard/Sites/example/vendor/phing/phing/bin/../classes:/Users/rickard/Sites/example/vendor/pear/archive_tar:/Users/rickard/Sites/example/vendor/pear/console_getopt:/Users/rickard/Sites/example/vendor/pear/net_url2:/Users/rickard/Sites/example/vendor/pear/pear-core-minimal/src:/Users/rickard/Sites/example/vendor/pear/pear_exception:/Users/rickard/Sites/example/vendor/pear/versioncontrol_git:/Users/rickard/Sites/example/vendor/phing/phing/classes:.:/usr/local/Cellar/php/8.1.9/share/php/pear') [line 92 of /Users/rickard/Sites/example/vendor/phing/phing/classes/phing/tasks/ext/HttpTask.php]
```

byrond:
@agentrickard it looks like we need to update the version of `config_ignore` that is added with Composer during `drupal-first-install` of `the-build` to the dev version that supports Drupal 10 (like https://github.com/palantirnet/drupal-skeleton/pull/139). I can create a PR for that, which will need to be merged (along with cutting a new release of the-build) before this will pass.

Since the contrib modules are installed by the-build during `drupal-first-install`, we may not need to have them in the skeleton's `composer.json`. See https://github.com/palantirnet/drupal-skeleton/pull/138

byrond:
The Build doesn't specify version numbers in `drupal-first-install`, which is nice, because we don't have to update them. I wonder if we need to temporarily change the `minimum-stability` to allow the `dev` version of Config Ignore to be selected and installed.

The alternative would be to temporarily remove Config Ignore until there is a stable release for Drupal 10.

byrond:
The `HTTP/Request2` errors look like a bug in Phing 2, which is no longer maintained. Phing 3 switches to using Guzzle. We use Request2 in the Acquia tasks. I think the Phing error is safe to ignore (as it also happens on `develop`), and we should probably rollback the update to Request2 until we are able to test with the Acquia tasks. It seems like we've met the goal of this PR in fixing the Pantheon install issue.

byrond:
I confirmed the `HTTP/Request2` is benign. The Phing file in question tries to use `include_once` to load the nonexistent HTTP/Request2.php file. It uses `@` to try and suppress errors, and that might be what is no longer working. Request2 is actually loaded by the Composer autoloader since it is in the `require` section of `the-build`.

I confirmed that the `settings.pantheon.php` is indeed downloaded and copied to the `sites/default` directory as expected in spite of the PHP error.

To move forward, we should:
* remove the update to Request2
* complete the remaining tasks in the description of this PR

byrond:
I tested this again from scratch with the testing steps I added to the description and still got the following error:
`Unable to load Drupal settings. Check your --root, --uri, etc.`

I wasn't able to run the installer without starting DDEV first, because the installer relies on DDEV to check for a database connection and run the first Drupal install.

I think the change to `install.xml` is necessary, but it just doesn't resolve this issue. This will likely only be resolved by a documentation update once we find the right steps to install for Pantheon sites.

byrond:
If I wait to `Uncomment the lines for Pantheon, and comment out the lines for Acquia` until after running `the-build-installer`, it completes successfully. However, `ddev restart` throws an error about the database type not matching, which requires `ddev delete`. The, after `ddev start`, I can run `phing install`, and the site works as expected using the `web` webroot and Pantheon settings.

byrond:
@agentrickard I believe the issue is here:
https://github.com/palantirnet/the-build/blob/035f00045cc461713dc8b57fd56e7c183792911f/targets/install.xml#L147-L148

When we set the `docroot` property in `.ddev/config.yaml` and start DDEV, it creates `web/sites/default/settings.ddev.php` and a `settings.php` to include it. Then, the lines mentioned above during install remove the `web` directory and these settings files in order to move `docroot` to `web`. This isn't a problem if the files already exist in `docroot` or if we aren't trying to switch to `web`.

byrond:
@agentrickard The update I made to the installer works for me. I updated the testing steps in the description for what to verify.

PR #211, opened 2023-01-23
TECH-33 Copy phpcs.xml to different path depending on host

iajon:
## Overview
Modifies `phpcs.xml` install location for pantheon builds.

[Jira ticket](https://palantir.atlassian.net/browse/TECH-33)

## Testing
I tested this using the following steps:
1) `git clone https://github.com/palantirnet/drupal-skeleton.git`
2) `cd drupal-skeleton`
3) `composer install`
4) `cd vendor/palantirnet && rm -rf the-build`
5) `git clone https://github.com/palantirnet/the-build.git`
6) `cd the-build`
7) `git checkout --track origin/TECH-33-update-phpcs-xml-during-the-build-install`
8) `ddev start`
9) `ddev ssh`
10) `vendor/bin/the-build-installer`
11) When prompted, specify `pantheon` as the hosting platform.
12) Once the install is complete, ensure that `phpcs.xml` has the following lines:

```
  <!-- Directories to scan. -->
  <file>web/modules/custom</file>
  <file>web/themes/custom</file>
```

13) Repeat steps 1-10
14) When prompted, specify `acquia` as the hosting platform.
15) Once the install is complete, ensure that `phpcs.xml` has the following lines:

```
  <!-- Directories to scan. -->
  <file>docroot/modules/custom</file>
  <file>docroot/themes/custom</file>
```


byrond:
I don't think I adequately described what we need here. We actually need to modify the contents of:
https://github.com/palantirnet/the-build/blob/develop/defaults/install/phpcs.xml

Specifically, we need to change these lines in the file when the web root is changed from `docroot` to `web`:
https://github.com/palantirnet/the-build/blob/fc7b0ff3fa16ea30c2e5b0c23c5f623bfb744db6/defaults/install/phpcs.xml#L32-L33

The file should always be copied to the project root.

See the following example:
https://github.com/palantirnet/the-build/blob/develop/targets/install.xml#L164-L173

In the above example, we are updating the `installer-paths` in `composer.json` whenever we change the web root. See:
https://github.com/palantirnet/drupal-skeleton/blob/develop/composer.json#L82-L91

byrond (changes requested):
See comments.

byrond (approved):
Perfect! Thank you!

PR #212, opened 2023-01-25
Update the-build to use headless chrome

byrond:
* Updates the installed copy of `behat.yml` to work with Headless Chrome. This is being worked on and tested with https://github.com/palantirnet/drupal-skeleton/pull/141
* Fixes an issue where The Build will not install properly when DDEV is not in use (on CircleCI, for example). The existence of `settings.ddev.php` is now checked before attempting to copy it. (See `targets/install.xml`)
* Switches from `Behat\MinkExtension` to `Drupal\MinkExtension`, which is the proper configuration according to [their documentation](https://behat-drupal-extension.readthedocs.io/en/v4.0.1/localinstall.html). This is mostly relevant when using the "wait for AJAX to finish" step. See https://github.com/jhedstrom/drupalextension/blob/93bf8cd/src/Drupal/DrupalExtension/Context/MinkContext.php#L159 and https://github.com/pantheon-systems/example-drops-8-composer/issues/400

### DIFF
--- a/targets/install.xml
+++ b/targets/install.xml
@@ -142,11 +142,15 @@
                 <!-- The one that's available is the location we're moving from. -->
                 <available file="${build.dir}/web/core" type="dir" property="drupal.root.old" value="web" />
                 <available file="${build.dir}/docroot/core" type="dir" property="drupal.root.old" value="docroot" />
-                <available file="${build.dir}/${drupal.root.old}/sites/default/settings.ddev.php" type="file" property="ddev_settings_exists" value="true" />
+                <available file="${build.dir}/${drupal.root}/sites/default/settings.ddev.php" type="file" property="ddev_settings_exists" value="true" />
+                <available file="${build.dir}/${drupal.root.old}/sites/default/settings.ddev.php" type="file" property="old_ddev_settings_exists" value="true" />
 
                 <!-- Copy settings.ddev.php if it doesn't exist in the old drupal root. -->
                 <if>
-                    <not><equals arg1="${ddev_settings_exists}" arg2="1" /></not>
+                    <and>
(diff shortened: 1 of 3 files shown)


targets/install.xml
```diff
@@ -142,11 +142,15 @@
                 <!-- The one that's available is the location we're moving from. -->
                 <available file="${build.dir}/web/core" type="dir" property="drupal.root.old" value="web" />
                 <available file="${build.dir}/docroot/core" type="dir" property="drupal.root.old" value="docroot" />
-                <available file="${build.dir}/${drupal.root.old}/sites/default/settings.ddev.php" type="file" property="ddev_settings_exists" value="true" />
+                <available file="${build.dir}/${drupal.root}/sites/default/settings.ddev.php" type="file" property="ddev_settings_exists" value="true" />
```
agentrickard, line 145: @byrond What's up with these unrelated changes?

agentrickard (commented):
Not sure why we updated install.xml here.

targets/install.xml
byrond: > @byrond What's up with these unrelated changes?

@agentrickard see description:
> Fixes an issue where The Build will not install properly when DDEV is not in use (on CircleCI, for example). The existence of settings.ddev.php is now checked before attempting to copy it. (See targets/install.xml)

This was causing a build failure on https://github.com/palantirnet/drupal-skeleton/pull/141: 
https://app.circleci.com/pipelines/github/palantirnet/drupal-skeleton/183/workflows/531485e2-2665-466c-b422-8a9cd5219680/jobs/604

byrond: In hindsight, I could have created a separate PR for that change, but it was blocking this one, so I opted to fix it here.

agentrickard (approved):

PR #216, opened 2023-02-03
Load headless chrome during circle builds

byrond:
Fixes an issue that came up after https://github.com/palantirnet/drupal-skeleton/pull/141 was merged into the skeleton. It now uses headless Chrome, so we need to update our CircleCI config here to load the browser on the correct port.


byrond:
This is included in https://github.com/palantirnet/the-build/pull/211

PR #218, opened 2023-02-28
Move Acquia DB backups to cloud v2.

acidaniel:
https://palantir.atlassian.net/browse/DEV-32

## Test Intructions:

* Add this to an existing project and run `composer install`
* Ensure in the `build.yml` from the existing project you have the following variables:
   
   ```
  # Configuration for targets/acquia.xml
  acquia:
    backups: 'artifacts/backups'
    site: 'palantirnet'
    database: 'palantirnet'
    env: prod (default one if not defined)
   ```
   
  * Ensure there is a `.env` with acquia clould credentials specified in the 1Password
  
  * Run `ddev restart`
  
  * Run `ddev phing acquia-get-backup`
  
  * Ensure backup was downloaded inside the backups directory specified in `backups`
  
  * Run `ddev phing drupal-load-db` and ensure database dump downloaded is imported correctly.


composer.json
```diff
@@ -23,7 +23,8 @@
         "pear/versioncontrol_git": "@dev",
         "phing/phing": "^2.14",
         "phpmd/phpmd": "^2.4",
-        "phpspec/prophecy-phpunit": "^2"
+        "phpspec/prophecy-phpunit": "^2",
+        "typhonius/acquia-php-sdk-v2": "^2.0"
```
jesconstantine: @acidaniel I think Bec was interested in not adding this package to `the-build` -- I wonder if there's a way to only add it if a developer chooses "Acquia" as the hosting provider during setup?

I agree that it makes development on top of the API pretty seamless and I've benefited from it on DHS.

jesconstantine: Maybe adding it as a suggested package would work to begin with? https://getcomposer.org/doc/04-schema.md#suggest

```json
{
    "suggest": {
        "typhonius/acquia-php-sdk-v2": "Needed to use the phing AcquiaTask"
    }
}
```

There are apparently some paths forward for programmatically moving suggestions from `suggest` to `require`: https://stackoverflow.com/questions/27212257/is-there-a-way-to-interactively-install-suggested-composer-packages

Might be something to think about!

src/TheBuild/Acquia/GetLatestBackupTask.php
```diff
@@ -90,207 +55,97 @@ class GetLatestBackupTask extends AcquiaTask {
...
-    $json = $this->getApiResponseBody("/sites/{$this->realm}:{$this->site}/envs/{$this->env}/dbs/{$this->database}/backups.json");
-
-    $writer = new \FileWriter($backups_file);
-    $writer->write($json);
+   * Get latest backup from specified environment.
+   */
+  protected function getLatestBackup($client, $environment_uuid) {
+    $backup = new DatabaseBackups($client);
+    $backups = $backup->getAll($environment_uuid, $this->database);
+    $filepath = $this->dir . '/' . $this->env . '_' . $this->database . '_' . 'sql.gz';
```
jesconstantine: @acidaniel would it make sense to also set the file to load from backups based on this value?

https://github.com/palantirnet/the-build/pull/218/files#diff-339072f52845d316656969cbac2815305c0ccea0ce9de789ba73d78849336067L141-L145

composer.json
jesconstantine: @acidaniel One last follow-up based on our slack conversation:

> If we made it suggested, could we add a task that checks for the dependency and provides helpful output on failure? Then we could make any of the acquia tasks be dependent on that one.

src/TheBuild/Acquia/GetLatestBackupTask.php
```diff
@@ -361,10 +196,15 @@ protected function validate() {
     if (empty($this->database)) {
       $this->database = $this->site;
     }
+    // Check .env file exists.
+    if (!file_exists(".env")) {
+      throw new \BuildException(".env file is needed in the root of project with credentials see .env.example.");
```
jesconstantine: @acidaniel should there be an `.env.example` as part of this PR?

composer.json
acidaniel: @jesconstantine I have added the `typhonius/acquia-php-sdk-v2` to the `suggest` key, and also I have updated the task `setup-host` that will include the package to the main `composer.lock` of the project and add the package only if `acquia` is selected as a hosting.

src/TheBuild/Acquia/GetLatestBackupTask.php
acidaniel: @jesconstantine The load from backups task will load any `*sql.gz` file that exists in the `artifact/backups` directory regardless the name.

src/TheBuild/Acquia/GetLatestBackupTask.php
acidaniel: @jesconstantine Yes, I just added a task to copy this file from the defaults.

acidaniel:
@jesconstantine This is ready to be reviewed now. Since the package was removed now I'm getting issues in CircleCI, do you think we need to skip this file to be linted?

targets/install.xml
```diff
@@ -281,8 +281,15 @@
                 <!-- Site-level setup -->
                 <!-- @see https://docs.acquia.com/cloud-platform/performance/memcached/enable/ -->
                 <httpget url="https://docs.acquia.com/_downloads/cloud-memcache-d8+.php" dir="${build.dir}/${drupal.root}/sites/${drupal.site.dir}" filename="settings.acquia-memcache.php"/>
+                <!-- Installing Cloud API SDK v2. -->
+                <composer command="suggest" composer="${composer.composer}">
+                  <arg line="--working-dir ${build.thebuild.dir}" />
+                  <arg value="--list" />
+                  <arg line="| xargs printf -- '%s:^2.0' | xargs composer require --working-dir ${application.startdir}" />
+                </composer>
```
jesconstantine: Oh this is neat! 

How does this know which suggested packages to install?

🤔 If we have the ability to run composer from a phing task, I wonder if we should just install the package with `composer require` here instead of adding it to the composer.json as `suggest` and running this `suggest` command.  

What do you think?

targets/install.xml
```diff
@@ -281,8 +281,15 @@
...
+                <!-- Copy .env.example file to store Acquia Cloud Credentials -->
+                <copy file="${build.thebuild.dir}/defaults/install/.env.example" todir="${application.startdir}" overwrite="true"/>
```
jesconstantine: I know the `.env.example` currently only has Acquia data in it, but I could see it being useful for non-acquia hosted projects too, what do you think?

Should we copy the file somewhere more globally and add the platform specific data to it here instead?

src/TheBuild/Acquia/GetLatestBackupTask.php
jesconstantine: @acidaniel I think that's my concern.  I think manual backups have a different file patterns but could get loaded instead of the one that we download from prod.  To me this is the same issue as what we were discussing in slack [here](https://palantirnet.slack.com/archives/CCVCR9W3A/p1677778984570949?thread_ts=1677688552.129409&cid=CCVCR9W3A) - but am I think about it correctly?

I added this to our `build.yml` to ensure the correct backup was loaded: https://github.com/palantirnet/wisdhs-public/pull/566/files#diff-0fad5e53ffd9ff1b0af504ff6c4d0961af03f34af69d9e00de527f2f29c9cb2bR50-R56

composer.json
jesconstantine: Great! I just added a follow up question in a comment at that part of the code: https://github.com/palantirnet/the-build/pull/218#discussion_r1128195681

jesconstantine:
@acidaniel you asked:

> Since the package was removed now I'm getting issues in CircleCI, do you think we need to skip this file to be linted?

Can we either:

* ensure the task that installs the acquia cloud API SDK package is run before running the tests?
* exclude `src/Acquia` [here](https://github.com/palantirnet/the-build/blob/develop/.circleci/config.yml#L47-L52), then run the new phing task to install the acquia cloud API SDK package, and then run through the code review commands specifically on `src/Acquia`

What do you think?

targets/install.xml
acidaniel: I added this as `suggest` since this was your suggestion, I was having lot of issues with specifying the version though, this is why I ended with this `xargs` solution, I saw there is something similar already when the user specify `platform.sh` as hosting [here](https://github.com/palantirnet/the-build/blob/develop/targets/install.xml#L303-L306). If we want to keep things simpler then this is how it should be.

composer.json
acidaniel: @jesconstantine I have removed this from suggested and I have added this directly to a `composer require` IF the acquia provider is selected as a hosting vendor.

src/TheBuild/Acquia/GetLatestBackupTask.php
acidaniel: @jesconstantine I would love to see your commit, but I got 404 message :'( (you know why. _Rub salt in the wound_).

acidaniel: @jesconstantine I just changed here the extension, I had a typo, please let me know if there something else to do here.

acidaniel: @jesconstantine I think I found a way to improve performance by 96% of loading database with only using Linux commands. Here is a bit of background:

1. Using `drupal-load-db` was using a `gunzip -c` command that when I run it it took around **10 minutes** to finish the import
2. Using `<exec executable="ddev"...` took around **56 seconds** to load, but this is not an option since we need to ensure we are outside the container. So this option is discarded.
3. Figured out and reading that by using native Linux commands like `gzip -dc` in combination with a direct `mysql` command takes only **46 seconds**

You can see how I am implementing it really straightforward. So following the general formula about performance improvement in percentage which is:

```
( new - old ) / old * 100%
```

We have:

```
(42 - 600)/600*100= -93
```

What do you think about this?

jesconstantine: I'm so sorry about posting that link, I wasn't thinking. 😦 

Here's the snippet: 
```yml
    build:
        # Configuration for the database loading utility.
      load_db:
        # Pattern to match gzipped dump files
        export_pattern: "./artifacts/*.sql.gz"
        file: "./artifacts/dhsinternet_d9_db_backup.sql.gz"
```

targets/install.xml
jesconstantine: Thanks and sorry for the bad suggestion, I didn't realize we could use composer require in phing but it makes sense that we could!  I'm up for keeping it simple and matching existing patterns. 👍🏼

defaults.yml
```diff
@@ -110,7 +113,7 @@ drupal:
       # Database connection for Drupal.
       database:
         # This value is required, but a default is set in the-build.xml.
-        # database: drupal
+        database: db
```
byrond: Do we need to change the default in the-build.xml instead of overriding it by default here?

defaults.yml
```diff
@@ -134,9 +137,9 @@ drupal:
         #   $> gunzip -c FILENAME.sql.gz | drush sqlc
         #
         # Command to extract text contents of the backup file.
-        contents_command: gunzip -c
+        contents_command: gzip -dc
         # Command to load database contents into Drupal.
-        mysql_command: drush sqlc
+        mysql_command: mysql --host=${drupal.site.database.host} --user=${drupal.site.database.username} --password=${drupal.site.database.password} --database=${drupal.site.database.database}
```
byrond, line 143: I don't think `drupal-load-db` could be run outside DDEV before, but we have an opportunity here to make this work like Drush and Behat (where it calls DDEV if those Phing commands are called outside the container). See https://github.com/palantirnet/the-build/pull/180/files.

We could tackle that in a separate issue if we aren't causing a regression here by changing the `mysql_command`.

defaults.yml
```diff
@@ -134,9 +137,9 @@ drupal:
         #   $> gunzip -c FILENAME.sql.gz | drush sqlc
         #
         # Command to extract text contents of the backup file.
-        contents_command: gunzip -c
+        contents_command: gzip -dc
```
byrond, line 141: When we talked earlier on the DevOps call, it sounded like this change wasn't necessary to improve performance (or was it)?

byrond:
I confirmed the testing steps work on an existing project using this branch. I think we just need to address the issues mentioned above, and this will be good to go.

defaults.yml
```diff
@@ -15,6 +15,9 @@ build:
   # The destination host, either 'acquia', 'pantheon', 'platformsh', or 'other'. This is
   # currently only used when setting up the settings.php file for a Drupal site.
   host: acquia
+    # Adding specific packages for specific host. One per line.
+    host_packages:
+      - "typhonius/acquia-php-sdk-v2:^2.0"
```
byrond, line 21: This does not get added by the install script if we choose Acquia.

byrond: This issue has been fixed.

src/TheBuild/Acquia/GetLatestBackupTask.php
byrond: This is checking for the `.env` file in the root, but it actually needs to exist in the `.ddev` directory for the environment variables to be created. I'm going to remove this, because the API already checks for the presence of these variables and fails accordingly:
```
example > acquia-get-backup:

[getacquiabackup] Couldn't find ACQUIA_CLOUD_API_KEY env variable.
[getacquiabackup] Couldn't find ACQUIA_CLOUD_API_SECRET env variable.

BUILD FAILED
```

defaults.yml
byrond: Instead of setting a default value here for the database name (which could break multisite behavior), I updated the default for the first database that gets copied with `defaults/install/the-build/build.yml`. See https://github.com/palantirnet/the-build/pull/218/commits/90b7d7db6829c0a127fd9cdada31b9f0de9ed1a9

Now, when a multisite is added, and that default is not present, it defaults to the site directory name (as it did before this PR).

byrond:
This was successfully tested on another client project and ready to merge.

byrond (approved):

patrickfweston (approved):

PR #227, opened 2023-08-16
DEV-41 - Pin circle php image to 8.1.18.

raylwalters:
Pin circle php image to 8.1.18 to deal with phpcs out of memory bug on 8.1.19. 

https://palantir.atlassian.net/browse/DEV-41

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     working_directory: ~/project
     docker:
-      - image: cimg/php:8.1-browsers
+      - image: cimg/php:8.1.18-browsers
       - image: cimg/mysql:5.7
         command: --max_allowed_packet=16M
         environment:


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   build:
     working_directory: ~/project
     docker:
-      - image: cimg/php:8.1-browsers
+      - image: cimg/php:8.1.18-browsers
```
raylwalters, line 9: This can be reverted once the phpcs error is fixed, https://github.com/php/php-src/issues/11972

becw (approved):

.circleci/config.yml
byrond: @raylwalters @becw based on the linked PHP issue, should we have pinned to 8.1.17?

raylwalters: I thought we were having issues with 8.1.19. I have PNET pinned to 8.1.18 and haven't been seeing the errors.

Girgias:
If you are having a memory leak in 8.1.19 this is unlikely to be because of the linked issue.

Please open a new issue on php-src

PR #230, opened 2023-10-27
DEV-25: Error reporting on "dirty directory" failures

jaymejohnson:
This PR adds extra output when a build fails due to a dirty artifact that lists the changed files. 

To test: We will need to artificially create a dirty artifact scenario. Manually changing a composer.lock file to e.g. update 
"plugin-api-version": "2.3.0" to "plugin-api-version": "2.6.0" did the job on the DHS project. 

### Testing steps
1. In a project, checkout to a new testing branch that runs on circle.
2. Run `composer update palantirnet/the-build:dev-dev-25-dirty-artifact-reporting` and commit it.
3. Make a manual edit to composer.lock to trigger a dirty repo error.
4. commit it in whatever manner the project requires in order to trigger the deploy step and the artifact build. 

**Note:** When I was testing this initially I ran into an issue where it was never getting to the deploy step because circle would kill the build when the dirty repository checks were run in the step up step. If that happens, you can just comment out the lines in your project's `.circleci/config.yml` that are running the dirty repository checks. 

5. Once CircleCI runs the artifact build we expect a visible fail on the deploy step, as well as details in the test target output about what files were modified that caused the artifact to be not-safe-to-build. What you see should look something like this:
Screenshot 2023-11-16 at 6 29 50 PM

Link to example: https://app.circleci.com/pipelines/github/palantirnet/wisdhs-public/6655/workflows/72668daf-1094-4a61-bd7e-5d0723b5416f/jobs/16061


targets/artifact.xml
```diff
@@ -346,5 +347,5 @@
         <exec dir="${artifact.directory}" command="git branch -D ${artifact.git.temporary_branch}" checkreturn="true" logoutput="true" />
     </target>
 
-
+
```
byrond: Remove spaces.

byrond (commented):
This looks great, @jaymejohnson! I viewed the CircleCI pipeline where this was tested, and the output looks good.

Once the spacing change on line 350 is reverted, this will be good to go!

targets/artifact.xml
jaymejohnson: All set!

becw (approved):
Wahoo!

PR #232, opened 2023-11-17
unpin phpmd version

byrond:
Versions of phpmd greater than 2.13 require the `--suffixes` flag. We pinned to 2.13 until we figured this out. This removes the constraint and allows 2.14 and later to be installed while fixing the calls to `phpmd` in the-build.


becw (approved):
Looks good! The updated usage of the phpmd command works under phpmd 2.13 AND 2.14, nifty.

PR #234, opened 2024-01-23
DEV-30 - Move phpmd config to phpmd.xml.

raylwalters:
User story: [DEV-30: Move phpmd config into phpmd.xml](https://palantir.atlassian.net/browse/DEV-30)

### Description

This moves the phpmd config in to a separate phpmd.xml file.

### Testing instructions

1. Update the build for this branch in a project 
2. Run the build installer `vendor/bin/the-build-installer`
3. Run code review `vendor/bin/phing test` to verify that phpmd runs.


becw:
When we move `defaults/standard/phpmd.xml` to `defaults/install/phpmd.xml`, we'll need to make sure to note this in the update steps. We should also consider bumping by a minor version (rather than a patch version) when we release this change.

jaymejohnson (approved):
LGTM

PR #237, opened 2024-08-16
Drupal11

byrond:
User story: [DEV-67: Update drupal-skeleton to Drupal 11](https://palantir.atlassian.net/browse/DEV-67)

- Updates dependencies for Drupal 11 compatibility.
- Updates `the_build_utility` module to be compatible with D11.

This is being tested on https://github.com/palantirnet/drupal-skeleton/pull/159


byrond:
Tests are now passing with the changes from https://github.com/palantirnet/the-build/pull/238. While an update to Node 20 or 22 and Yarn 4 would be prudent, it is no longer blocking the D11 update.

byrond:
I created https://palantir.atlassian.net/browse/DEV-73 before noticing it only affects Drupal 11. Since the error is with the `standard` profile, we might be able to remove it entirely rather than switch to `minimal` with our Phing task.

See https://www.drupal.org/node/3432357

This will be fixed by https://github.com/palantirnet/the-build/pull/239

raylwalters (approved):
Tested in the drupal-skeleton pr and this works as expected.

raylwalters (approved):
Looks good.

PR #238, opened 2024-08-23
DEV-70: Update install node dependencies task in the build

byrond:
- removes test for installed dependencies, as Yarn already does that
- removes call to install dependencies within Drupal core directory
- removes option to resolve dependencies relative to Drupal core directory

This has a [companion PR](https://github.com/palantirnet/drupal-skeleton/pull/160) for drupal-skeleton, which adds the required dependencies to the project `package.json`. Tests passed on CircleCI, and the change to use that version of the skeleton was reverted: https://app.circleci.com/pipelines/github/palantirnet/the-build/589/workflows/fced6b6f-4a86-4e1a-8728-fd3610a1d741/jobs/600

See testing steps on the companion PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,7 +72,7 @@ jobs:
       # Create a new project using the drupal-skeleton project
       - run:
           name: Create a new Drupal project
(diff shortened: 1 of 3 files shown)


becw (commented):
👍 to this change. I made one suggestion by way of a commit (caching yarn packages on CircleCI).

becw (approved):

PR #239, opened 2024-08-23
DEV-74: Remove standard profile after installing the-build

byrond:
As of Drupal 10.3, profiles can be uninstalled!

https://www.drupal.org/node/3432357 

This changes `drupal-first-install` to uninstall the `standard` profile instead of calling `drupal-change-profile`.

**Open Questions**

Do we need to keep `drupal-change-profile` in The Build, or should we remove it?


becw:
I love this, and I am 👍👍👍👍 (more thumbs than I personally possess) on removing `drupal-change-profile` from the-build.

We may want to specify that `phing drupal-first-install` will now require Drupal 10.3 or newer, although `palantirnet/the-build` as a whole does not yet.

becw (approved):
Love the # of deleted lines, works as expected during testing.

PR #246, opened 2025-05-21
Don't check .env files into artifacts

byrond:
`.env` files often contain sensitive keys or secrets used by environments. Don't check these into artifact builds if they exist on local dev environments during `phing artifact`.

### DIFF
--- a/defaults/artifact/gitignore
+++ b/defaults/artifact/gitignore
@@ -1,3 +1,4 @@
 # This .gitignore file is specific to the artifact.
 settings.local.php
 node_modules
+.env


becw (approved):
Simple!

becw:
I think that the artifact should only copy over files that are checked into the source repository, but if for some reason this was getting into an artifact... it shouldn't! If a project is running an older version of the-build and needs to work around this, you can configure a custom `.gitignore` for the artifact in `.the-build/build.yml`:

```
artifact:
  # Path to a template .gitignore file to use in the artifact.
  gitignore_template: "${build.thebuild.dir}/defaults/artifact/gitignore"
```

becw:
The current test failure isn't related to this change -- it looks like an issue introduced when we updated the yarn and nodejs versions required in the drupal-skeleton.

